PR #4, opened 2020-08-27
Epad8 824 wide template tweaks

azinck:


services/drupal/web/themes/epa_theme/templates/content/node--full.html.twig
```diff
@@ -15,7 +15,11 @@
 
 {% set header_info %}
   {% if sidenav %}
-    {{ node.group_content|view({type: 'parent_group_label_formatter', settings: {link: false}}) }}
+    {% if node.field_wide_template.value %}
+      {{ node.group_content|view({type: 'group_homepage_node_formatter', settings: {link: true}}) }}
+    {% else %}
+      {{ node.group_content|view({type: 'group_homepage_node_formatter', settings: {link: false}}) }}
+    {% endif %}
```
kmonahan, line 22: If it's just the value of the link setting that changes based on the field_wide_template, could you do something like 
`{{ node.group_content|view({type: 'group_homepage_node_formatter', settings: {link: node.field_wide_template.value is not empty}}) }}` (or however you need to determine that there is no valid value)?

services/drupal/web/themes/epa_theme/templates/content/node--full.html.twig
```diff
@@ -45,7 +49,17 @@
...
 
+{% set display_sidenav = false %}
+
 {% if sidenav|render|striptags|trim %}
+  {% set display_sidenav = true %}
+{% endif %}
+
+{% if node.field_wide_template.value %}
+  {% set display_sidenav = false %}
+{% endif %}
```
kmonahan, line 60: Any reason not to combine these two? `{% if sidenav|render|striptags|trim and not node.field_wide_template.value %} // set it to true`?

PR #13, opened 2020-09-04
[EPAD8-856] Display entity title or link text for link lists

briantschu:
A new field formatter which extends `LinkFormatter` with the same options. When an internal entity is selected as the link and there is no text in the link text sub-field, the entity's label will be used as the link text.

### DIFF
--- a/services/drupal/web/modules/custom/epa_links/src/Plugin/Field/FieldFormatter/LinkWithEntityTitleOrLinkTextFormatter.php
+++ b/services/drupal/web/modules/custom/epa_links/src/Plugin/Field/FieldFormatter/LinkWithEntityTitleOrLinkTextFormatter.php
@@ -1,0 +1,145 @@
+<?php
+
+namespace Drupal\epa_links\Plugin\Field\FieldFormatter;
+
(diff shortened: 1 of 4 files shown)


services/drupal/web/modules/custom/epa_links/src/Plugin/Field/FieldFormatter/LinkWithEntityTitleOrLinkTextFormatter.php
```diff
@@ -0,0 +1,158 @@
...
+
+      // If the url is internal, try to get the linked entity's title.
+      if (empty($settings['url_only']) && $url->isRouted()) {
+        $url_params = $url->getRouteParameters();
+        $entity_type = key($url_params);
+        $entity = $this->entityTypeManager->getStorage($entity_type)->load($url_params[$entity_type]);
+        if ($entity && $entity->label()) {
+          $link_title = $entity->label();
+        }
+      }
```
azinck, line 107: Instead of duplicating everything in this method...couldn't we just do this little bit of logic to modify the $items list and inject the correct title, then call the parent method and let it handle the hard work?

PR #28, opened 2020-09-14
Updated gov't site alert with new text

cherrypj:
The problem is that the updated text includes an SVG icon, which is too large compared to the rest of the text. I'm hoping @dcmouyard can review it. 

epadd8-912


azinck:
@dcmouyard Have you had a chance to look at this?

dcmouyard:
@azinck I reviewed this in the last sprint. This PR needs to wait until we update USWDS, which we now have a PR for in #51.

PR #40, opened 2020-09-17
EPAD8-744: Update node templates to use web area short name

dcmouyard:


services/drupal/web/themes/epa_theme/templates/content/node--full.html.twig
```diff
@@ -16,9 +16,9 @@
 {% set header_info %}
   {% if sidenav %}
     {% if node.field_wide_template.value %}
-      {{ node.group_content|view({type: 'group_homepage_node_formatter', settings: {link: true}}) }}
+      {{ node.group_content|view({type: 'parent_group_label_formatter', settings: {link: true}}) }}
```
azinck, line 19: I think this is going to end up linking to the group rather than the homepage for the group.  I guess I need to update our group_homepage_node_formatter.

PR #44, opened 2020-09-21
Update pattern-lab configuration

cherrypj:
Hide "All" navigation option
Add US EPA logo
Set "HTML" to open by default (defaultPatternInfoPanelCode)


dcmouyard (approved):
I created a light & dark version of the EPA logo for Pattern Lab since the SVG on the production site will go away once this new D8 site is deployed.

PR #48, opened 2020-09-23
Allow load balancer to respond to other hostnames

gustavderdrache:
This PR changes the `alb-hostname` variable into an optional list of strings for other host names to match against incoming requests.

### DIFF
--- a/infrastructure/terraform/alb.tf
+++ b/infrastructure/terraform/alb.tf
@@ -87,10 +87,7 @@ resource "aws_lb_listener_rule" "frontend_https" {
 
   condition {
     host_header {
-      # If the alb-hostname var isn't set, use the site hostname (this matters in cases
-      # where we're behind a CDN and the ALB's domain name is different from the
-      # public-facing domain name.)
-      values = [var.alb-hostname == null ? var.site-hostname : var.alb-hostname]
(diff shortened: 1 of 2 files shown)


infrastructure/terraform/alb.tf
```diff
@@ -87,10 +87,7 @@ resource "aws_lb_listener_rule" "frontend_https" {
 
   condition {
     host_header {
-      # If the alb-hostname var isn't set, use the site hostname (this matters in cases
-      # where we're behind a CDN and the ALB's domain name is different from the
-      # public-facing domain name.)
-      values = [var.alb-hostname == null ? var.site-hostname : var.alb-hostname]
+      values = concat([var.site-hostname], var.alb-hostnames)
```
azinck, line 90: @gustavderdrache I just want to be sure this is right -- so we're setting the host header here to an array of all possible hostnames for the site?  Is that really what we want?

gustavderdrache: No, this is the load balancer's matching condition: we're allowing any of these host headers. The host is still fixed in the nginx config [here](https://github.com/USEPA/webcms/blob/stable/services/drupal/default.conf#L48).

PR #61, opened 2020-10-02
Epad8 730 paramaterize email further

azinck:
I'm parameterizing the port and protocol (Gopa asked for this).  I also provided a new setting to disable the sending of workflow notifications so that we could allow emailing to work in general without having to have all those emails get sent.  Please take a look and let me know what you think.

(And yes, the --debug flag for cron snuck in there but I'm not bothering to refactor).


gustavderdrache (approved):
Make sure to update our Parameter Store values before this is merged or else the Terraform run will fail.

PR #63, opened 2020-10-05
added link to our box standard

caedstrom:


caedstrom:
@dcmouyard this is the kind of change, we would like to be able to make to pattern lab. Did we submit this correctly?

dcmouyard (approved):
@caedstrom @cherrypj Looks great! 👍

cherrypj:
@caedstrom : I think we'll have to add related links and other documentation/how-tos to the actual examples. Like for the box, above, we would add the "Related link: Web Standards Boxes" to the text in the example boxes.

box-move-text

PR #72, opened 2020-10-06
EPAD8-897: External image links

dcmouyard:


kmonahan (approved):
👍 Don't think I've seen too many sites put an external link indicator on images. It's a nice touch.

PR #103, opened 2020-10-26
add link to USWD color page

caedstrom:
@dcmouyard  I added a link to USWD color page. I added it just after the title because I thought it would get lost at the bottom of the page.


cherrypj:
Not sure we can use the `show_documentation` variable here

cherrypj (approved):

dcmouyard (approved):
👍

PR #110, opened 2020-10-26
Add styling to is-active main-menu link

cherrypj:
@dcmouyard : we noticed that there was no indicator of the active page in the main menu


dcmouyard (approved):
👍

PR #127, opened 2020-11-03
EPAD8-960: Add Colorbox CSS library

dcmouyard:


dcmouyard:
@kmonahan This PR just adds the Colorbox CSS. I originally wanted it in the `css/lib` directory, but I think one of our Gulp tasks kept deleting it, perhaps `compileStyles`?

For future reference, do you know if there’s a way to change that?

PR #159, opened 2020-11-18
Renamed accordion dir to _accordion

cherrypj:
This effectively hides it from Pattern Lab. Is that true @dcmouyard ?


cherrypj:
Note that we do not want to remove accordions from the site (as they are in use on the search facet pages, like /newsreleases/search). But we do want to remove them from the Pattern Lab navigation.

dcmouyard:
> This effectively hides it from Pattern Lab. Is that true @dcmouyard ?

@cherrypj It does, but we also need to update any Drupal templates that point to the accordion component.

azinck:
@dcmouyard Where does this stand?  Can we either push it through to get committed or close the PR?

PR #160, opened 2020-11-19
[EPAD8-1022] Use fid for file migration high water mark

mrclay:
Timestamps allowed some duplicates to be missed.

https://forumone.atlassian.net/browse/EPAD8-1022

### DIFF
--- a/services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaFile.php
+++ b/services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaFile.php
@@ -44,13 +44,12 @@ class EpaFile extends File {
   public function query() {
     $query = parent::query();
 
-    // Don't use timestamp highwater because it isn't unique, and stops/restarts
-    // can result in files being skipped. Remove the parent's sort and replace
-    // with fid.
+    // Remove core's timestamp sorting because this causes problems mixing
+    // with our use of fid high water mark. We don't use timestamp high water
+    // because it isn't unique, and stops/restarts can result in files being
+    // skipped.
     $sort =& $query->getOrderBy();
     unset($sort['f.timestamp']);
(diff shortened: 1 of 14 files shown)


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaFile.php
```diff
@@ -38,6 +38,23 @@ protected function initializeIterator() {
...
+  public function query() {
+    $query = parent::query();
+
+    // Don't use timestamp highwater because it isn't unique, and stops/restarts
+    // can result in files being skipped. Remove the parent's sort and replace
+    // with fid.
+    $sort =& $query->getOrderBy();
+    unset($sort['f.timestamp']);
+    // Use fid highwater mark
+    $query->orderBy('f.fid');
```
azinck, line 53: This is not necessary -- there's other code that handles adding the orderBy for the high water mark.  We only need to remove the timestamp sort.

services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaFileEntityItem.php
```diff
@@ -0,0 +1,38 @@
...
+  public function query() {
+    $query = parent::query();
+
+    // Don't use timestamp highwater because it isn't unique, and stops/restarts
+    // can result in files being skipped. Remove the parent's sort and replace
+    // with fid.
+    $sort =& $query->getOrderBy();
+    unset($sort['f.timestamp']);
+    // Use fid highwater mark
+    $query->orderBy('f.fid');
```
azinck, line 33: Same as the other file; this is not necessary -- there's other code that handles adding the orderBy for the high water mark.  We only need to remove the timestamp sort.

azinck (changes requested):

PR #169, opened 2020-11-30
[EPAD8-987] Be pickier about external link style

mrclay:
In page zip exports, the current page href will just be an HTML
filename and--lacking ".gov"--it would normally get the external link
styling. This ensures that doesn't happen, at least for exported nodes.

https://forumone.atlassian.net/browse/EPAD8-987


azinck:
I put some feedback on EPAD8-987 of some improvements we need to make here.

PR #171, opened 2020-12-02
Allow drush-migrate.sh to be gracefully stopped

mrclay:


services/drupal/scripts/ecs/drush-migrate.sh
```diff
@@ -42,13 +43,30 @@ run_migration() {
       --limit=$batch_size \
       --continue-on-failure \
       "$migration"
+
+    # Check if we've asked to halt. Migration's own status field isn't really reliable
+    # for discerning whether a migration has been stopped or completed.
+    halted=$(drush state-get epa.migrations_halted)
```
gustavderdrache, line 49: Make sure you quote the capture here.
```suggestion
    halted="$(drush state-get epa.migrations_halted)"
```

services/drupal/scripts/ecs/drush-migrate.sh
```diff
@@ -42,13 +43,30 @@ run_migration() {
...
   done
 
+  # Calculate the difference in seconds for some timing statistics
   finish="$(date +%s)"
+  time=$((finish - start))
 
   # Determine how many unprocessed items were left behind by the migration.
   unprocessed="$(drush ms "$migration" --field=unprocessed)"
 
+  if test -n "$halted"; then
```
gustavderdrache, line 63: What's the reasoning for the second check here? Why not return directly in the loop?

services/drupal/scripts/ecs/drush-migrate.sh
mrclay: Ah because the captured string could have spaces...

services/drupal/scripts/ecs/drush-migrate.sh
mrclay: I wanted to output the unprocessed.

PR #176, opened 2020-12-04
Fix high water mark on our banner migration to ensure all get migrated

azinck:
Please take a look at this and tell me what you think.  We missed some records in the upgrade_d7_node_web_area_paragraph_banner_slide migration due to the fact that we used the revision ID as the high water mark.  Revision ID isn't unique since there can be multiple field values attached to a single revision with different deltas.  This change adds an expression that concatenates revision ID and delta to achieve a truly unique value to be used as the Highwater mark.

### DIFF
--- a/services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaMultivalueField.php
+++ b/services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaMultivalueField.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\epa_migrations\Plugin\migrate\source;
 
-use Drupal\migrate\Row;
+use Drupal\migrate\Plugin\MigrateIdMapInterface;
 use Drupal\migrate_drupal\Plugin\migrate\source\DrupalSqlBase;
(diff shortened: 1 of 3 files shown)


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaMultivalueField.php
```diff
@@ -35,6 +35,9 @@ public function query() {
       $query = $this->select($field_table, 'fd')
         ->fields('fd');
 
+      // Add an expression to let us generate a unique field for use as high water mark.
+      $query->addExpression('CONCAT(fd.revision_id, "-", fd.delta)', 'unique_id');
```
mrclay, line 39: SqlBase uses `$conditions->condition($high_water_field, $high_water, '>')` and MySQL doesn't allow comparison with a computed column: `Unknown column 'unique_id' in 'where clause'`. So I think our plugin will have to override initializeIterator() and use `> CONCAT(...`.

mrclay (changes requested):

PR #183, opened 2020-12-11
[EPAD8-1051] Make classname rewrites safe and remove some replacements

mrclay:
The 2nd commit fixes a clear bug, but I could just fix the str_replace is desired.


azinck:
Please refactor this to be agnostic of the order in which the classnames appear on the wrapper element.

PR #188, opened 2020-12-17
Keep record of allowed 1 unprocessed revision

mrclay:
This is just to note that we're allowing this 1 failure and we should make sure this is in the DB before next migration.

### DIFF
--- a/scripts/allow-unprocessed.sh
+++ b/scripts/allow-unprocessed.sh
@@ -15,7 +15,7 @@ started_by="bschumacher"
 # e.g. drush state-set epa.allowed_unprocessed.upgrade_d7_node_revision_document 1
 
 script=$(cat <<EOF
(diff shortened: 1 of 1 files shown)


scripts/allow-unprocessed.sh
```diff
@@ -15,7 +15,7 @@ started_by="bschumacher"
 # e.g. drush state-set epa.allowed_unprocessed.upgrade_d7_node_revision_document 1
 
 script=$(cat <<EOF
-
+  drush state-set epa.allowed_unprocessed.upgrade_d7_node_revision_page_panelizer 1
```
azinck, line 18: I actually don't think we want to do this...I don't think we'd expect this to happen next time we run the migration (hopefully we don't run out of memory and have to restart it).

PR #189, opened 2020-12-17
EPAD8-1078 use proper dom manipulation classes in order to only add t…

azinck:
…he attributes we actually have data for

### DIFF
--- a/services/drupal/web/modules/custom/epa_migrations/src/EpaMediaWysiwygTransformTrait.php
+++ b/services/drupal/web/modules/custom/epa_migrations/src/EpaMediaWysiwygTransformTrait.php
@@ -39,16 +39,6 @@ trait EpaMediaWysiwygTransformTrait {
 
     $pattern = '/\[\[(?<tag_info>.+?"type":"media".+?)\]\]/s';
 
-    $media_embed_replacement_template = <<<'TEMPLATE'
-<drupal-media
-  alt="%s"
-  data-align="%s"
-  data-caption="%s"
-  data-entity-type="media"
-  data-entity-uuid="%s"
-  data-view-mode="%s"></drupal-media>
-TEMPLATE;
-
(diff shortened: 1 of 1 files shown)


services/drupal/web/modules/custom/epa_migrations/src/EpaMediaWysiwygTransformTrait.php
```diff
@@ -77,14 +67,27 @@ public function transformWysiwyg($wysiwyg_content, EntityTypeManager $entityType
           );
         }
         else {
+          $doc = new DOMDocument();
```
mrclay, line 70: Good idea

services/drupal/web/modules/custom/epa_migrations/src/EpaMediaWysiwygTransformTrait.php
```diff
@@ -77,14 +67,27 @@ public function transformWysiwyg($wysiwyg_content, EntityTypeManager $entityType
...
-            htmlentities(stripslashes(urldecode($tag_info['fields']['field_caption[und][0][value]']))) ?? '',
-            $media_entity_uuid,
-            $view_modes[$tag_info['view_mode']]
-          );
+          $el->setAttribute('data-align',$alignment);
+
+          $caption = htmlentities(stripslashes(urldecode($tag_info['fields']['field_caption[und][0][value]']))) ?? '';
+          if (!empty($caption)) {
+            $el->setAttribute('data-caption',$caption);
+          }
```
mrclay, line 82:
```suggestion
          $caption = stripslashes(urldecode($tag_info['fields']['field_caption[und][0][value]'] ?? ''));
          if (!empty($caption)) {
            $el->setAttribute('data-caption', $caption);
          }
```

This does two things: moves `??` expression so notice won't occur calling urldecode, and removes the now unnecessary htmlentities.

mrclay (changes requested):

PR #190, opened 2020-12-18
Mark group content as ignored in migration if entity not found

mrclay:


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaOgMembership.php
```diff
@@ -78,19 +80,18 @@ public function prepareRow(Row $row) {
 
     // Load the D8 entity.
     $entity = $this->entityTypeManager->getStorage($type)->load($etid);
+    if (!$entity) {
+      // Entity can't be found, so ignore row
+      return $this->ignoreRow($row, "etid '$etid' failed to load");
```
azinck, line 85:
```suggestion
      return $this->ignoreRow($row, "Unable to migrate group relationship due to missing '$type' entity with ID '$etid'.");
```

PR #191, opened 2020-12-18
Match media blocks so that multiple blocks can't be matched as 1

mrclay:
We match `[[{ ... }]]` THEN validate that it's a media block.


mrclay:
I need to test this one better.

azinck:
I'm going to close this PR since I'm not convinced we need to act on this.  Can revisit if ill-effects are found pertaining to media token matching.

PR #207, opened 2021-01-16
[EPAD8-986] Document fixes; revision update

briantschu:


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/process/EpaSetLatestRevision.php
```diff
@@ -140,7 +140,7 @@ public function transform($value, MigrateExecutableInterface $migrate_executable
     // current revision (as obtained from D7), retaining its same state.
     //
     // If they do match, skip this step and goto forward revision logic.
-    if ($node->vid !== $d7_vid) {
+    if ($node->get('vid')->value !== $d7_vid) {
```
azinck, line 143: I'm surprised we need this change...did $node->vid really not behave as expected?

services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/process/EpaSetLatestRevision.php
```diff
@@ -163,7 +163,9 @@ public function transform($value, MigrateExecutableInterface $migrate_executable
 
       $newer_draft_revision = $this->getHeaviestDraftRevision($d7_forward_revisions);
 
-      if ($newer_draft_revision->vid < $d7_vid) {
+      // If the newer draft revision has a vid lower than the current revision,
+      // we want to give it a new vid so it is set as the latest revision.
+      if ($newer_draft_revision->vid < $node->get('vid')->value) {
```
azinck, line 168: I'm similarly confused by the syntax choice here for accessing $node->vid, but I also think the logic here might be wrong.  Just above here, on lines 143-157, we've potentially created a new revision in order to set the current revision correctly.  Now, on line 166, we need to be considering whether the vid that we have determined *should* be our latest revision actually *is* our latest revision.  At the very least, we should be comparing $newer_draft_revision->vid to $new_current_revision->vid (if $new_current_revision exists)...but I think even that is not bulletproof.  I think what we should actually be doing is calling \Drupal\content_moderation\ModerationInformationInterface::getLatestRevision to load what D8 knows to be the latest revision for this node and comparing its value to D7's $newer_draft_revision.  If they don't match, then we need to re-save the correct vid in D8.

azinck (changes requested):
Let me know if these make sense to you.

services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/process/EpaSetLatestRevision.php
briantschu: `$node->vid` is returning a `FieldListItem` object. I found an entity API method for getting the vid and implemented it in my latest change: `$node->getRevisionId()`.

services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/process/EpaSetLatestRevision.php
briantschu: Thanks for noting this. I agree we should compare to what D8 knows to be the latest revision. `getLatestRevision` was replaced with a core function. I implemented with: `$node_storage->getLatestRevisionId($node->id());`

azinck:
This looks good to me, but I'm waiting until I have the chance to do a little testing on the stage environment with the existing code in place (I'm trying to reproduce the negative counts on some of our migrations) before merging this.

PR #218, opened 2021-01-22
Update overview.twig

caedstrom:
@cherrypj and @dcmouyard The EPA Web Standards link is broken. Also the intranet drupal training page link seems correct but doesn't work for me either. -ce


cherrypj:
I don't know that it's broken--I can click through. (I think it's better to leave it on the same domain (rather than hard-coding the www).) Rather, the URL opens inside the Pattern Lab iframe. So maybe it needs a target attribute to bust it out.

azinck:
@dcmouyard Are we ever planning on merging this? Or is this no longer relevant?

cherrypj:
No longer relevant

PR #226, opened 2021-02-04
[EPAD8] Migration config cruft

briantschu:
When I uninstalled and re-installed the epa migrations module locally, these config changes were detected. I split them into a separate PR because I'm not familiar with the `use_og_table` source setting and wanted to make sure this is a desired change.

### DIFF
--- a/services/drupal/config/sync/migrate_plus.migration.upgrade_d7_group_content_file.yml
+++ b/services/drupal/config/sync/migrate_plus.migration.upgrade_d7_group_content_file.yml
@@ -6,7 +6,7 @@ dependencies:
     module:
       - epa_migrations
 _core:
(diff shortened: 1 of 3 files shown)


services/drupal/config/sync/migrate_plus.migration.upgrade_d7_group_content_file.yml
```diff
@@ -22,7 +22,6 @@ source:
     name: etid
   d7_entity_type: file
   d8_entity_type: media
-  use_og_table: true
```
azinck, line 25: Actually this change needs to go the other direction.  I made the change to this file only, neglecting to change it in the module's original config.  I haven't really found a lot of value in trying to maintain the source config from the module, TBH.

PR #228, opened 2021-02-04
[EPAD8-1125] Disable webform ip address collection

briantschu:


services/drupal/web/modules/custom/epa_forms/epa_forms.module
```diff
@@ -101,6 +102,12 @@ function epa_forms_form_alter(&$form, FormStateInterface $form_state, $form_id)
         '#markup' => t('This action will also remove any associated webform entity and <strong>delete its submissions</strong>.'),
       ];
       break;
+
+    case 'webform_settings_submissions_form':
+      if (isset($form['submission_behaviors']['form_remote_addr'])) {
+        unset($form['submission_behaviors']['form_remote_addr']);
+      }
+      break;
```
azinck, line 110: I'm thinking $form['submission_behaviors']['form_remote_addr']['#access'] = FALSE; would be more in line with the standard way to handle this rather than just un-setting it entirely (and risk breaking other code that expects this element to exist).

briantschu: Got it! I will update.

PR #237, opened 2021-02-12
Clean up migration scripts

briantschu:
Remove the `--continue-on-failure` flag as it is no longer needed. The migration should stop at the first sign of any individual migration failing.

Remove the paragraph migration testing script as it is no longer needed.


azinck:
These are probably right, I just don't fully understand these changes, so let's chat when you get back in the office.

PR #239, opened 2021-02-13
[EPAD8-1141] Migrate media keywords to media_tags vocabulary

briantschu:


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaFileEntityItem.php
```diff
@@ -34,4 +34,43 @@ public function query() {
     return $query;
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public function prepareRow($row) {
```
azinck, line 40: Is all this necessary?  I seem to recall from my experience with Migrate eons ago that there's a way to just push a string into a taxonomy field and it'll just auto-match the values to existing terms in the vocabulary and generate new ones as-needed.  Have you looked into this at all?

azinck: In fact, from a brief look it seems maybe the entity_generate is designed to do just this very thing.  If so, don't we just need to extract a string of term names?

briantschu: In the migration configuration we are using entity_generate. The problem is, the source plugin only has the Term ID, since that is what is stored in the field_keywords on the file entity. This code is looking up the name for the terms so we can generate them. I'm returning an array of tids and term names here so we can be sure our terms aren't going to steal an ID from a term that will be migrated later.

PR #261, opened 2021-03-24
[EPAD8-1187] Update faq topics field label and form display

briantschu:


azinck:
I left a note on the ticket about switching to the simple HS widget.

PR #262, opened 2021-03-24
[EPAD8-1192] Preprocess regulation link form elements to remove external link help

briantschu:


azinck:
I ended up committing a slightly different approach in epa_core_field_widget_form_alter()

PR #268, opened 2021-03-31
[EPAD8-1195] Add svgxuse polyfill and add some IE11 fallbacks

kmonahan:


dcmouyard (changes requested):
Try mixin override instead.

dcmouyard (approved):

PR #280, opened 2021-04-07
[EPAD8-1092] Switch to visually hidden text on facets and icons

kmonahan:


services/drupal/web/themes/epa_theme/source/_patterns/05-components/icon/icon.twig
```diff
@@ -29,15 +29,17 @@
...
         <use xlink:href="{{ epa_theme_image_path }}sprite.artifact.svg#{{ icon }}"></use>
       </svg>
       {{ icon }}
     </p>
   {% endfor %}
 {%- else -%}
   {% spaceless %}
-    <svg class="{{ classes }}"{% if is_hidden %} aria-hidden="true"{% endif %}{% if label %} aria-label="{{ label }}"{% endif %}>
+    <svg class="{{ classes }}"{% if is_hidden %} aria-hidden="true"{% endif %}{% if label %} role="img"{% endif %}>
+      {% if label %}<title>{{ label }}</title>{% endif %}
```
dcmouyard, line 42: @kmonahan Is the switch to `<title>` for the label to get this work in NVDA?

kmonahan: Yep, exactly. It's an older (I think) method of adding accessible alt-like text to SVGs, which I switched to here since NVDA wouldn't read aria-label, even with a role in place, in Lysne's testing.

PR #281, opened 2021-04-07
[EPAD8-1094] Add alert label to boxes for screen readers

kmonahan:


services/drupal/web/themes/epa_theme/source/_patterns/05-components/box/box.twig
```diff
@@ -28,6 +28,9 @@
 {% endset %}
 
 <div class="{{ classes }}"{% if role %} role="{{ role }}"{% endif %}>
+  {%- if label -%}
+    <div class="u-visually-hidden">{{ label }}</div>
+  {%- endif -%}
   {% if title.text %}
     <{{ title.tag }} class="box__title">
       {% if title.url %}
```
dcmouyard, line 36: @kmonahan I’m leaning towards moving the accessible label to inside `.box__title` to help users who navigate by headings. What do you think?

kmonahan: The reason I didn't put it there initially was because it looked like title was optional. But I could set it up so it's part of the title if the title is present and in its current location if not.

dcmouyard (approved):
👍

PR #282, opened 2021-04-07
[EPAD8-1095] Add text filter to label warning text

kmonahan:


azinck:
@kmonahan @dcmouyard This might be the right general approach, but there are 2 things I don't like about it:
1) It means we're running additional regex on every single bit of rich text we output.  This is fairly resource-intensive both in terms of CPU and RAM. 
2) We're [parsing HTML in regex](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454)

I recognize that EPAFilterEmptyParagraphs is using regex for a similar purpose.  It should be refactored. But it's also had a lot of testing done with it in place so I'm not too paranoid about it right now.  I don't want to pile more technical debt on, so if we need to do HTML parsing and manipulation we should be using tools appropriate for that (like DOMDocument).

How vital is it that we inject this warning text?  I'm fairly anxious about introducing code like this this late in the game, code that will literally touch and process and potentially manipulate every single piece of text on the entire site.

kmonahan:
Ah, I modeled this after EPAFilterEmptyParagraphs. I can take a pass at refactoring next week.

dcmouyard:
Closing this since we don’t want to merge this now. We may revisit this after site launch.

PR #285, opened 2021-04-09
Epad8 1184 group menu change

azinck:
This should not be merged until the completion of Migration 7 due to the fact that it will break all earlier databases.


azinck:
We need to wait to merge this until all environments running the 'main' branch are receiving the migration 7 (or higher) DB.  That's because it will break any environments running older DBs. The main environment I'm concerned about here is the old (SAIC) beta.epa.gov.site which is unlikely to ever see the migration 7 DB.  So we need to ensure EPA has a new environment up and running and ready to receive the newer migration DB before we break their beta site.

PR #286, opened 2021-04-09
Epad8 1233 migrate authmap

azinck:
This code is meant to add entries to the authmap table to establish the correct mapping between Drupal and SAML users.  You can test this by running the user migration and ensuring that entries are created in the authmap table.


briantschu (approved):
I made one update to pass the value of the 'name' instead of the FieldItemList object.

I think you'll want to update the base of this PR to main, not migration.

azinck:
Thanks for the fix!

I'm making this against migration because if it gets committed to main first, then it's going to break EPA's beta site.  The mappings need to be migrated before the other config can be used.

azinck:
This should go in before migration 8 is started.

PR #291, opened 2021-04-15
[EPAD8-1243] Account for scrollbar in full-width elements

kmonahan:


dcmouyard (approved):
@kmonahan Neat trick to use JS set the scrollbar CSS variable! 👍

PR #297, opened 2021-04-22
Apply pull quote style to blockquotes generally

cherrypj:
@dcmouyard : is this the right way to make all `blockquote`s look like pull quotes? We find the default `blockquote` styling to be a little jarring, and we're finding that people have been using `blockquote`s a lot.


dcmouyard (changes requested):
@cherrypj We generally want to leave out element selectors in components to prevent collisions with other styles.

My preferred approach would be to move these pull-quote styles to the blockquote element styles in base, then extend those styles to this component.

I can push up this change shortly.

dcmouyard (approved):

PR #299, opened 2021-04-23
[EPAD8-1255] Transform grid columns with 'size-NofX' classes

briantschu:


azinck:
I'll be honest that I don't feel well-equipped to review this, but let's give it a try.

PR #305, opened 2021-04-30
[EPAD8-1251] Remove the alter hook from media_migration

briantschu:


azinck:
How about we do this via hook_module_implements_alter rather than a patch, @briantschu

Any reason that wouldn't work here?

PR #308, opened 2021-05-20
[EPAD8-1308] Body content missing

briantschu:


services/drupal/web/modules/custom/epa_migrations/src/Plugin/migrate/source/EpaPanelizerNode.php
```diff
@@ -105,7 +105,20 @@ public function prepareRow(Row $row) {
...
+      // Fetch the node_content pane for the body field.
+      $body_pane = $this->select('panels_pane', 'pp')
+        ->fields('pp')
+        ->condition('pp.did', $did)
+        ->condition('pp.type', 'node_content')
+        ->orderBy('pp.position')
+        ->orderBy('pp.panel')
+        ->execute()
+        ->fetchAll();
+      $row->setSourceProperty('body_pane', $body_pane);
```
azinck, line 117: Perhaps I'm missing something, but it appears that we're using this to extract pane info and inject it into the 'body_pane' field, which we're then, in our migration, reading into our field_paragraphs field.  The problem I see is that this pane itself doesn't directly contain any data at all.  It's just a placeholder that tells Panelizer to emit the body content.  What we really need to inject into our new field_paragraphs field is the actual existing body content from the D7 node. Is that happening somewhere and I'm just missing it?

briantschu: The `body_pane` source data contains the Node Content pane(s) we're passing through the `epa_panes_to_paragraphs` process plugin. The process plugin generates the paragraphs that are ultimately referenced by `field_paragraphs`. This is the same process plugin we're using for non-panelizer content migrations to assign paragraphs to `field_paragraphs`.

After reviewing the non-panelizer configuration for `field_paragraphs`, I realized the config updates I made in this PR  deviate a bit. I'm not passing the generated paragraphs through the sub_process plugin. The sub_process step would accommodate the case of there being multiple paragraphs generated by `epa_panes_to_paragraphs`. While this was working on my test nodes without the sub_process step, I just tested adding that step locally and I get the expected result. I'll go ahead an add the sub_process step in to this PR so we're processing `field_paragraphs` consistently.

PR #309, opened 2021-05-21
Adding redirects to 301.map

cherrypj:
To edit, follow the normal approach of creating a feature branch based on main, make your edits in your branch, merge it to integration to test on dev, then take out a PR against main.

https://forumone.atlassian.net/browse/EPAD8-753


cherrypj:
I'm confident that these redirects will work OK, but I don't want to bring down the dev.byf1 site.

azinck:
Please see my comments in Jira for the changes that need to be made.

PR #321, opened 2021-06-14
Add redirects from the existing D7 site

cherrypj:
Here you go @azinck


azinck:
Feedback left on the Jira ticket.

PR #345, opened 2021-07-09
Epad8 1352 slideshows larger

jbwagner:
Have the slides use the medium (640x360px) image instead of the small, update styles accordingly

### DIFF
--- a/services/drupal/config/sync/responsive_image.styles.responsive_gallery_image.yml
+++ b/services/drupal/config/sync/responsive_image.styles.responsive_gallery_image.yml
@@ -5,6 +5,8 @@ dependencies:
   config:
     - image.style.gallery_medium
     - image.style.gallery_small
+    - image.style.huge_gallery
+    - image.style.large_gallery
 id: responsive_gallery_image
 label: 'Responsive Gallery Image'
 image_style_mappings:
@@ -17,5 +19,7 @@ image_style_mappings:
       sizes_image_styles:
         - gallery_medium
         - gallery_small
+        - huge_gallery
+        - large_gallery
 breakpoint_group: responsive_image
 fallback_image_style: gallery_medium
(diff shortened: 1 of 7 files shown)


services/drupal/config/sync/responsive_image.styles.responsive_gallery_image.yml
```diff
@@ -13,9 +13,9 @@ image_style_mappings:
     multiplier: 1x
     image_mapping_type: sizes
     image_mapping:
-      sizes: '(max-width: 320px) calc(100vw - 2rem), 320px'
+      sizes: '(max-width: 640px) calc(100vw - 2rem), 640px'
       sizes_image_styles:
         - gallery_medium
         - gallery_small
```
dcmouyard, line 19: Here’s where we’ll add `gallery_large` and `gallery_huge`.

dcmouyard (changes requested):
@jbwagner  Great work!

I forgot to specify some details in the ticket: Since we’re changing the gallery responsive image style to 640px, we also want to add larger images so it works on high DPI devices.

We’ll want to create new Large Gallery (960x540) and Huge Gallery (1280x720) image styles and reference them in the responsive image style.

Also, could you investigate why we have nested figures in the markup of the gallery images? I think is a regression from when we first implemented the slideshow gallery component.

```
<figure class="figure figure--gallery">
  <div class="figure__media">
    <div class="contextual"></div>
    <figure class="figure figure--gallery"> // Can we remove this?
      <div class="figure__media"> // Can we remove this?
        <a>
          <img>
        </a>
      </div> // Can we remove this?
    </figure> // Can we remove this?
  </div>
  <figcaption class="figure__caption"></figcaption>
</figure>
```

jbwagner:
@dcmouyard Should be all set. The double <figure> was some odd templating, and it might have been renamed at some point. I left what was there but just added a stripped down media template for gallery images. Should only apply for this use case anyway so should be good.

Also just created and added the new image styles for the larger media. Matched those to the ones that already existed, so should slot in nicely. Tested this out on my local with test content.

dcmouyard (approved):

PR #350, opened 2021-07-13
Add size attribute to select element (multiple already allowed)

cherrypj:
https://forumone.atlassian.net/browse/EPAD8-1319


briantschu:
@cherrypj I see there's a merge conflict -- I wonder if you were using 'main' as the base for your feature branch locally or didn't have the latest config imported before making the change and exporting it. Please let me know if you'd like any help resolving the merge conflict.

cherrypj:
I did use main; I can't see how to force feature to be the base branch.

briantschu:
I think what you'll want to do is check out this branch locally and then `git rebase release`. Then you can `git push -f`. The `-f` is necessary because you'll be 'rewriting history' by changing the base of the branch. This isn't a problem since no one else has the branch checked out.

PR #360, opened 2021-07-21
EPAD8-1237 taxonomy dropdown label

jbwagner:


azinck:
Please update this to use the aria-label attribute on each select element. @jbwagner

PR #363, opened 2021-07-21
[EPAD8-1369] Enable the lateruntime processor

briantschu:
This PR actually enables the late runtime purge processor, instead of only enabling the module. When I saved the config and exported, the change to enable the cron processor came along for the ride as well.﻿


azinck:
The late runtime processor caused performance problems so we will not be using it.

PR #367, opened 2021-07-23
Fixed orcdizux

cherrypj:
We want to capture www.epa.gov/orcdizux/datafiles/batchfiles/ and send them to www3.epa.gov/otaq/datafiles/batchfiles/


cherrypj:
@azinck : can you push this nginx redirect change to ci-work? This way, we can push it to prod without also putting the EPAD8-1372 patch in (need that reviewed first).

azinck:
@cherrypj Done!

PR #377, opened 2021-07-29
[EPAD8-1349] Adding hook_update_N() to change "Document" media items that use json or xml files to become "Other" media bundle

carsoncho:
Note: I haven't tested this against a large or prod-like database, but wanted to get this in front of some eyes sooner rather than later. If there's ways I can optimize the queries please let me know.

### DIFF
--- a/services/drupal/web/modules/custom/epa_media/epa_media.install
+++ b/services/drupal/web/modules/custom/epa_media/epa_media.install
@@ -12,29 +12,21 @@ function epa_media_update_9000() {
 function epa_media_update_9001(&$sandbox) {
   $database = \Drupal::database();
 
-  if (!isset($sandbox['total'])) {
-    // Find all "Document" media entities that have a .json or .xml file.
-    $media_ids = $database
-      ->query("SELECT m.entity_id
+  // Find all "Document" media entities that have a .json or .xml file.
+  $media_ids = $database
+    ->query("SELECT m.entity_id
         FROM media__field_media_file AS m
         INNER JOIN file_managed AS fm ON m.field_media_file_target_id = fm.fid
         WHERE fm.filename like '%.json' OR fm.filename like '%.xml'
         AND m.bundle = 'document';")
-      ->fetchCol('entity_id');
+    ->fetchCol('entity_id');
 
-    $sandbox['total'] = count($media_ids);
-    $sandbox['current'] = 0;
-    $sandbox['offset'] = 0;
-  }
+  $sandbox['total'] = count($media_ids);
 
   // Do not continue if no media items are found.
   if (empty($sandbox['total'])) {
-    $sandbox['#finished'] = 1;
     return t('No Media entities to be processed.');
   }
-
-  // Arbitrary limit to process 50 at a time.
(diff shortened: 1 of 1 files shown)


services/drupal/web/modules/custom/epa_media/epa_media.install
```diff
@@ -5,3 +5,131 @@
...
+  // Do not continue if no media items are found.
+  if (empty($sandbox['total'])) {
+    $sandbox['#finished'] = 1;
+    return t('No Media entities to be processed.');
+  }
+
+  // Arbitrary limit to process 50 at a time.
+  $ids = array_slice($media_ids, $sandbox['current'], 50);
+
+  foreach ($ids as $id) {
```
azinck, line 42: I think we can avoid actually iterating over each record and each of these queries could be rewritten with IN criteria that matches on a subquery.   See an example below.

services/drupal/web/modules/custom/epa_media/epa_media.install
```diff
@@ -5,3 +5,131 @@
...
+        INNER JOIN {media_field_data} AS mfd ON m.mid = mfd.mid
+        INNER JOIN {media_revision} AS mr ON m.mid = mr.mid
+        INNER JOIN {media_field_revision} AS mfr ON m.mid = mfr.mid
+        SET
+	          m.bundle = 'other',
+	          mfd.bundle = 'other',
+	          mr.bundle = 'other',
+	          mfr.bundle = 'other'
+        WHERE m.mid = :id", [':id' => $id]);
+
```
azinck, line 55: This could be re-written to be something like this in order to do this update as a single query (not tested, so please test):
```
 $database
       ->query("UPDATE {media} AS m
         INNER JOIN {media_field_data} AS mfd ON m.mid = mfd.mid
         INNER JOIN {media_revision} AS mr ON m.mid = mr.mid
         INNER JOIN {media_field_revision} AS mfr ON m.mid = mfr.mid
         SET
 	          m.bundle = 'other',
 	          mfd.bundle = 'other',
 	          mr.bundle = 'other',
 	          mfr.bundle = 'other'
         WHERE m.mid IN (
             SELECT entity_id
             FROM {media__field_media_file} AS fmf 
               LEFT JOIN {file_managed} AS fm ON fm.fid = fmf.field_media_file_target_id 
             WHERE filename like '%.json' OR filename like '%.xml'
               AND bundle = 'document')");    
```

You should be able to follow a similar pattern for the rest of the queries, too.

services/drupal/web/modules/custom/epa_media/epa_media.install
carsoncho: I've now updated all queries to use arrays with IN and works as expected.

services/drupal/web/modules/custom/epa_media/epa_media.install
carsoncho: I've updated the queries to now use INs and passing 50 IDs at a time vs. individual queries. I'll test more today once I get the database working locally to see if there's any issue running all the IDs through it at once.

PR #378, opened 2021-07-30
EPAD8 1237 taxonomy dropdown aria-label and for attribute

jbwagner:


services/drupal/composer.json
```diff
@@ -259,7 +259,7 @@
         "drupal/search_api": "^1.19",
         "drupal/seckit": "^2.0",
         "drupal/shield": "1.x-dev",
-        "drupal/shs": "^1.0-alpha5",
+        "drupal/shs": "^1.0@alpha",
```
azinck, line 262: @jbwagner Is there a reason you made this change? It appears it doesn't result in a change to which version is actually resolved in composer.lock, so I'm very curious about this and whether it was intentional.

jbwagner: @azinck Wasn't specifically intentional, but that's what occurred when I tested if the patch applied over what we had. There shouldn't be any changes to the module, it showed the exact same version in Drupal on both accounts. I figured Composer knew what it was doing, so I left the change in. However, if it's something you think we shouldn't change, we just need to make sure the patch applies.

PR #379, opened 2021-08-02
[EPAD8-1377] Update core to 9.1.11

briantschu:
This PR updates core to 9.1.11, including dependencies. 

The update required an updated patch for the issue [Add Views EntityReference filter to be available for all entity reference fields](https://www.drupal.org/project/drupal/issues/2429699). We were using the patch from comment 389 and I've updated to the patch from comment 414.

### DIFF
--- a/services/drupal/composer.lock
+++ b/services/drupal/composer.lock
@@ -1602,24 +1602,23 @@
         },
         {
             "name": "doctrine/cache",
-            "version": "1.11.3",
+            "version": "1.12.1",
             "source": {
                 "type": "git",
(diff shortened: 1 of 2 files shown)


azinck:
This fails on a composer install with 
```
Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2021-02-12/2845144_50.patch

                                                                                                                                      
  [Exception]                                                                                                                         
  Cannot apply patch Allow referencing unpublished nodes if user has access to them (https://www.drupal.org/files/issues/2021-02-12/  
  2845144_50.patch)!                                                               
```

PR #409, opened 2021-08-19
[EPAD8-1353] Adding additional step in logic to ensure we're on a ContentEntityInterface before moving further into the code.

carsoncho:
The `GroupContent::loadByEntity()` expects a `ContentEntityInterface`. Adding this additional check ensures the method gets what it expects, resolving the error we're seeing.

```
...
$entity = $form_state->getformObject()->getEntity();
if ($entity->id()) {
  $group_contents = GroupContent::loadByEntity($entity);
  foreach ($group_contents as $group_content) {
    $group = $group_content->getGroup();
  }
}
...
```

### DIFF
--- a/services/drupal/web/modules/custom/epa_web_areas/epa_web_areas.module
+++ b/services/drupal/web/modules/custom/epa_web_areas/epa_web_areas.module
@@ -9,6 +9,7 @@ use Drupal\block\Entity\Block;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Entity\EntityForm;
 use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Url;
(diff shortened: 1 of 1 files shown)


services/drupal/web/modules/custom/epa_web_areas/epa_web_areas.module
```diff
@@ -140,6 +140,7 @@ function epa_web_areas_field_widget_form_alter($element, FormStateInterface $for
   if (!empty($plugin_type)
       && $plugin_type == 'media_library_widget'
       && $form_state->getFormObject() instanceof EntityForm
+      && $form_state->getFormObject()->getEntity() instanceOf ContentEntityInterface
```
azinck, line 143: @carsoncho I don't think we want to do it this way because it short circuits the other side of our conditional.  I think your approach here would fail on forms where we're creating a brand-new node and we have to try to extract the group context from the form_state storage.  Move your check down to the if statement around line 149.

carsoncho: @azinck Done. I've moved that instanceOf check down further in the logic per your recommendation.

PR #422, opened 2021-09-01
[EPAD8-1301] add tag instead of icon for external links

jbwagner:


dcmouyard (changes requested):
@jbwagner Great work! Just a few tweaks needed and some edge cases I forgot to mention in the ticket.

1. We want the exit tag inside the `<a>` so that it’s accessible to screen readers
2. If possible, tweak markup & styles to still get empty space between link text and tag
2. In `external-link--image.yml` and `external-link--inverse.yml`change the url variable to `https://example.net/`
3. Tweak the styles of the image and inverse variants of the external link component to look good with the exit tag and disable the current CSS approach if JS is enabled
4. Update all places that extend the external-link CSS to only work if JS is disabled (box and figure components plus their variants)
5. Update example content in Pattern Lab that uses `class="external-link"` to remove that class and use `href="https://example.net/"` instead of `#`.

services/drupal/web/themes/epa_theme/source/_patterns/05-components/external-link/_external-link.scss
```diff
@@ -22,11 +22,21 @@
   }
 }
 
-a#{$external-urls} {
-  @extend .external-link;
+html.no-js {
```
dcmouyard, line 26: @jbwagner This generates the following error in Stylelint: `Unexpected qualifying type selector   selector-no-qualifying-type`

[Stylelint](https://stylelint.io/) is used to lint Sass files. Warnings will break the build, so if you have a valid reason to break Stylelint rules you can have it ignore code in two ways:

1. Add `/* stylelint-disable-next-line */` to the line just before where the Stylelint warning is triggered.

2. To ignore several lines, add `/* stylelint-disable */` before the code in question and add `/* stylelint-enable */` afterwards.

The best way to fix the error here is to do the following:
```
/* stylelint-disable-next-line selector-no-qualifying-type */
html.no-js {
```

services/drupal/web/themes/epa_theme/js/src/external-links.es6.js
```diff
@@ -0,0 +1,20 @@
...
+  Drupal.behaviors.externalLinks = {
+    attach(context) {
+      const externalLinks = context.querySelectorAll(
+        "a:not([href=''], [href*='.gov'], [href*='.mil'], [href^='#'], [href^='?'], [href^='/'], [href^='.'], [href^='javascript:'], [href^='mailto:'], [href^='tel:'], [href*='webcms-uploads-dev.s3.amazonaws.com'], [href*='webcms-uploads-stage.s3.amazonaws.com'], [href*='webcms-uploads-prod.s3.amazonaws.com'])"
+      );
+      externalLinks.forEach(function(el) {
+        if (el.hasAttribute('href')) {
+          el.insertAdjacentHTML(
+            'beforeend',
+            `<span class="usa-tag external-link__tag">${Drupal.t('Exit')}</span>`
```
dcmouyard: This line was causing an error in prettier:
```
error  Replace `'Exit'` with `⏎··············'Exit'⏎············`  prettier/prettier
```

services/drupal/web/themes/epa_theme/source/_patterns/05-components/external-link/_external-link.scss
```diff
@@ -22,11 +22,21 @@
...
+  // Hide external link icon for anchors without an href attribute.
+  a:not([href])::after {
+    display: none !important;
+  }
+
+  .external-link__tag {
+    display: inline-block;
+    margin-left: 0.25em;
+    text-decoration: none;
+  }
```
dcmouyard: These lines were causing an error in the linter due to being indented too much.

dcmouyard (approved):
@jbwagner There were a few linting errors, but otherwise everything worked as expected. Since the sprint review is tomorrow I went ahead and fixed the linting issues.

PR #423, opened 2021-09-02
Attempting to fix epafiles redirect

john-moore-levesque:


azinck:
@cherrypj I've merged this to the integration branch for testing.  It's now deployed to http://qa.epa.byf1.io/ .  Since I don't have the context of what you were trying to fix with this change I'm not sure if it's been fixed correctly.  Can you test there to validate the fix before I merge this?

cherrypj:
It works great!

https://qa.epa.byf1.io/epafiles/something.txt redirects to https://www3.epa.gov/epafiles/something.txt, which is what we want.

Please do merge.

PR #432, opened 2021-09-09
EPAD8-1340 fix spanish metadata images

azinck:


cherrypj:
Yes! That's exactly right.

cherrypj (approved):

PR #440, opened 2021-09-16
[EPAD8-1408] remove back to top from keyboard flow

jbwagner:


services/drupal/web/themes/epa_theme/source/_patterns/05-components/back-to-top/back-to-top.twig
```diff
@@ -1,4 +1,4 @@
-<a href="#{{ top_element }}" class="back-to-top" title="{{ text }}" aria-hidden="true">
+<a href="#{{ top_element }}" class="back-to-top" title="{{ text }}" aria-hidden="true" tabindex="-1">
```
dcmouyard, line 1: This will permanently remove the back to top link from being tabbed to, which we don’t want. We should add the disabled attribute here instead, then remove it in JS when appropriate.

dcmouyard (changes requested):
@jbwagner We want to address this in JS so that the back to top link is accessible in all cases. I’ll make the changes necessary since tomorrow’s the sprint review.

dcmouyard (approved):

PR #448, opened 2021-09-22
Epad8 1449 update pattern lab for standalone page

cherrypj:
Hi @dcmouyard

I'm looking to reuse one of the existing PL page templates, but with different links and content. I'm uncertain if there are additional steps to take.

The branch is named badly, unfortunately, and doesn't accurately reflect the work here.


cherrypj:
The only change that should be in this PR is for update _page-default.twig (for updating the global footer)
SHA a62c6ecfa206c37523e52602f10e420f915497ac

Everything else was reverted in favor of a page in the WebCMS, where we can more easily update the content.

PR #451, opened 2021-09-22
Increase CPU back to 60% for high CPU usage tracking autoscale

cherrypj:


azinck:
@cherrypj If this is, in fact, the problem, I don't think this is the right way to fix it.  If there's a magic number of containers above which we can't scale due to IP reservation problems, then the max number of containers for the environment needs to be set low enough to avoid that.  We don't need to change the threshold setting.

azinck:
Closing this out.

PR #455, opened 2021-09-28
[EPAD8-1442] Published nodes missing transition dates

carsoncho:
I haven't merged this to integration yet. Wanted another set of eyes on this as a quick code review.

### DIFF
--- a/services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
+++ b/services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
@@ -86,17 +86,18 @@ function epa_workflow_update_9001(&$sandbox) {
 }
 
 /**
- * Update all nodes without field_review_deadline value to have value and a scheduled transition
+ * Update all nodes without field_review_deadline value to have value 1 year from now and a scheduled transition.
  */
 function epa_workflow_update_9002(&$sandbox) {
   // Set the review deadline value to a year from now
   $now = \Drupal\Core\Datetime\DrupalDateTime::createFromTimestamp(time());
-  $now->sub(new DateInterval('P1Y'));
+  $now->add(new DateInterval('P1Y'));
   $formatted = $now->format('Y-m-d\TH:i:s');
 
   if (!isset($sandbox['total'])) {
     // Query all published nodes that do not have a field_review_deadline value set.
     $result = \Drupal::entityQuery('node')
+      ->accessCheck(FALSE)
       ->condition('status', 1)
       ->notExists('field_review_deadline')
       ->execute();
@@ -110,8 +111,13 @@ function epa_workflow_update_9002(&$sandbox) {
   $nids = \Drupal::entityQuery('node')
     ->condition('status', 1)
     ->notExists('field_review_deadline')
-    ->range($sandbox['current'], $sandbox['current'] + $batch_size)
+    ->range($sandbox['current'], $batch_size)
     ->execute();
+
+  if (empty($nids)) {
+    $sandbox['#finished'] = 1;
+    return;
+  }
 
   /** @var \Drupal\node\NodeInterface[] $nodes */
   $nodes = \Drupal::entityTypeManager()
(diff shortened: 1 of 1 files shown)


services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+    $sandbox['total'] = count($result);
+    $sandbox['current'] = 0;
+  }
+
+  $batch_size = 25;
+
+  $nids = \Drupal::entityQuery('node')
+    ->condition('status', 1)
+    ->notExists('field_review_deadline')
+    ->range($sandbox['current'], $batch_size)
```
azinck, line 114: I believe this is going to cause a problem.  You're setting an offset which is going to result in skipping nodes because this query is run for each batch...and the act of processing the previous batch will prevent those previous nodes from appearing in this query when the next batch is run.  Basically, you don't need or want the offset here.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+  }
+
+  /** @var \Drupal\node\NodeInterface[] $nodes */
+  $nodes = \Drupal::entityTypeManager()
+    ->getStorage('node')
+    ->loadMultiple($nids);
+
+  foreach ($nodes as $node) {
+    $node->set('field_review_deadline', $formatted);
+    $node->save();
```
azinck, line 129: Have you tested 100% to be sure that this doesn't just result in a new draft revision being created, and it actually successfully updates the published revision?

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+  } else {
+    $sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);
+  }
+}
+
+/**
+ * Update all published nodes missing a transition date with a review deadline
+ * less than 3 weeks away or in the past.
+ */
+function epa_workflow_update_9003(&$sandbox) {
```
azinck, line 146: Any particular reason we're skipping _9002?  I don't really care, am just surprised.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+  // Create dates for 1 year in the future and 3 weeks in the future from now.
+  $now = \Drupal\Core\Datetime\DrupalDateTime::createFromTimestamp(time());
+  $one_year = $three_weeks = $now;
```
azinck, line 149: Unless there's handling in the DrupalDateTime class to avoid this, I believe this approach to assigning objects is going to result in $one_year and $three_weeks both referencing the exact same object, which is going to result in, I believe, the $formatted_three_weeks variable getting a date that is exactly one year and 3 weeks in the future.  Forgive me if I've missed something.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+    $sandbox['current'] = 0;
+  }
+
+  $batch_size = 25;
+
+  $nids = \Drupal::entityQuery('node')
+    ->condition('status', 1)
+    ->condition('field_review_deadline', $formatted_three_weeks, '<=')
+    ->notExists('field_scheduled_transition')
+    ->range($sandbox['current'], $batch_size)
```
azinck, line 176: Same problem as in _9001.  I don't think you want this offset.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+  $now = \Drupal\Core\Datetime\DrupalDateTime::createFromTimestamp(time());
+  $three_weeks = $now->add(new DateInterval('P3W'));
+  $formatted_three_weeks = $three_weeks->format('Y-m-d\TH:i:s');
+
+  if (!isset($sandbox['total'])) {
+    // Query all published nodes that don't have a scheduled transition date, but
+    // do have a review deadline
+    $result = \Drupal::entityQuery('node')
+      ->condition('status', 1)
+      ->condition('field_review_deadline', $formatted_three_weeks, '>=')
```
azinck, line 218: We shouldn't need or really want this condition.  _9003 should have handled catching everything that has a too-soon review date.  At this point we really just want to add a transition to *everything* that's published and missing a transition date.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+
+    $sandbox['total'] = count($result);
+    $sandbox['current'] = 0;
+  }
+
+  $batch_size = 25;
+
+  $nids = \Drupal::entityQuery('node')
+    ->condition('status', 1)
+    ->condition('field_review_deadline', $formatted_three_weeks, '>=')
```
azinck, line 230: Same comment as above.  Unnecessary.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+  }
+
+  /** @var \Drupal\node\NodeInterface[] $nodes */
+  $nodes = \Drupal::entityTypeManager()
+    ->getStorage('node')
+    ->loadMultiple($nids);
+
+  foreach ($nodes as $node) {
+    // Nothing needs to change on the node, but triggering a save will create a
+    // scheduled transition date 3 weeks prior via the EPAPublished service.
```
azinck, line 247: I assume you've validated this?  It's mildly surprising to me that it behaves this way, but is very convenient if true!

services/drupal/web/modules/custom/epa_workflow/epa_workflow.install
```diff
@@ -58,3 +58,202 @@ function epa_workflow_uninstall() {
...
+    $sandbox['current'] = 0;
+  }
+
+  $batch_size = 25;
+
+  $nids = \Drupal::entityQuery('node')
+    ->condition('status', 1)
+    ->condition('field_review_deadline', $formatted_three_weeks, '>=')
+    ->notExists('field_scheduled_transition')
+    ->range($sandbox['current'], $batch_size)
```
azinck, line 232: Again, no offset should be used here.

PR #500, opened 2021-11-02
[EPAD8-1526] Revision edit message

carsoncho:


services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -229,6 +230,28 @@ function epa_workflow_form_alter(array &$form, FormStateInterface $form_state, $
...
+  if ($node instanceof NodeInterface && !$node->isNew()) {
+
+    // Loading will always return the "Current revision".
+    /** @var \Drupal\node\NodeInterface $current */
+    $current = \Drupal::entityTypeManager()
+      ->getStorage('node')
+      ->load($node->id());
+
+    // The revision ID of the "Current revision" will be lower (older) than the most recent revision.
+    if ($node->getRevisionId() > $current->getRevisionId()) {
```
azinck, line 247: Minor quibble...but let's just check if these IDs differ at all.  It's true that the latest revision ID SHOULD always be higher than the current revision ID...but making that assumption provides us no benefit and I think we should just do a not-equals comparison here.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -229,6 +230,28 @@ function epa_workflow_form_alter(array &$form, FormStateInterface $form_state, $
...
+      $url = $node->toUrl('version-history')->toString();
+      $markup = Markup::create("<strong>Note:</strong> The revision you are editing is not the currently published content. It may include edits that are not visible on the published page. Refer to the <a target='_blank' href='$url'>revision page</a> to see other revisions and revert if necessary.");
```
azinck, line 249: I'm thinking we should use t() here rather than the Markup class.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
carsoncho: Adjusted to use not-equals comparison now.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
carsoncho: Removed usage of Markup and am now using the translation `t()` function as suggested.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -14,6 +14,7 @@ use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\RevisionLogInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\Markup;
```
azinck, line 17: Not needed anymore?

carsoncho: Removed.

azinck (approved):

PR #506, opened 2021-11-08
Add ALB to ECS cluster

gustavderdrache:
This PR adds an ALB behind the existing NLB. This is the first half of the migration away from Traefik; the second will be done after the ALB is deployed in order to minimize disruption. There is a new VPC security group required for the ALB. Its requirements are documented in the `terraform/network` module, and it has implications on the existing Drupal VPC security group.

### DIFF
--- a/terraform/infrastructure/load_balancer.tf
+++ b/terraform/infrastructure/load_balancer.tf
@@ -6,7 +6,7 @@ resource "aws_lb" "load_balancer" {
   subnets            = local.public_subnets
 
   access_logs {
-    bucket  = aws_s3_bucket.elb_logs.bucket
+    bucket  = coalesce(var.lb_logging_bucket, aws_s3_bucket.elb_logs.bucket)
     enabled = true
   }
 
@@ -81,7 +81,7 @@ resource "aws_lb" "app_load_balancer" {
   security_groups = [data.aws_ssm_parameter.alb_security_group]
 
   access_logs {
-    bucket  = aws_s3_bucket.elb_logs.bucket
+    bucket  = coalesce(var.lb_logging_bucket, aws_s3_bucket.elb_logs.bucket)
     enabled = true
   }
 
--- a/terraform/infrastructure/load_balancer.tf
+++ b/terraform/infrastructure/load_balancer.tf
@@ -12,7 +12,7 @@ resource "aws_lb" "load_balancer" {
 
   tags = var.tags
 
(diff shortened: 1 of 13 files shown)


terraform/network/security.tf
```diff
@@ -143,6 +154,40 @@ resource "aws_security_group_rule" "terraform_database_egress" {
...
+
+# Allow HTTP ingress to the ALB from arbitrary sources. Note that we don't limit
+# this to just the NLB's public subnets; in an AWS environment, the ALB's
+# security group rules apply not to the NLB's IP but the client IP.
+resource "aws_security_group_rule" "public_alb_http_ingress" {
+  description = "Allows HTTP ingress to the ALB"
+
+  security_group_id = aws_security_group.alb.id
+
+  type      = "ingres"
```
msglover, line 167: `type    = "ingress"`

terraform/infrastructure/load_balancer.tf
```diff
@@ -82,3 +71,86 @@ resource "aws_lb_target_group" "https" {
...
+resource "aws_lb" "app_load_balancer" {
+  name               = "webcms-${var.environment}-alb"
+  internal           = true
+  load_balancer_type = "application"
+
+  subnets         = local.public_subnets
+  security_groups = [data.aws_ssm_parameter.alb_security_group]
+
+  access_logs {
+    bucket  = aws_s3_bucket.elb_logs.bucket
```
msglover, line 84: Do we want to switch the logging over to the bucket provided by Task 7 with this commit?  The bucket is prefixed by the account number, so may be a little more involved.

gustavderdrache: That's a good idea. We don't have a similar bucket in our own AWS environment, so I think we'll probably want to continue creating the current S3 logging bucket for now.

However, as a transitional item, I'm going to create a new variable for the logging bucket and let it default to `null` – by default, it'll use the managed bucket, but you can override it to use the preconfigured bucket instead. We can probably move to fully unmanaged buckets in the same pass when we decommission the Traefik resources.

msglover: Do we want to wait on that variable before merging this?  If we're working on a deadline to get ALB's up then I'm comfortable merging and getting the variable change later

gustavderdrache: I just pushed the code. Sorry about the delay!

PR #518, opened 2021-11-18
Epad8 1449 update pattern lab menus

cherrypj:
I mistakenly reverted this change, but we do want it.


cherrypj:
@dcmouyard The footer file is already done; I'm not sure why it came in this commit too.

dcmouyard (approved):

PR #519, opened 2021-11-19
EPAD8 1494/1493 tables

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1493 https://forumone.atlassian.net/browse/EPAD8-1494


dcmouyard (changes requested):
@jbwagner Looks good! The only change needed is to tweak the mobile styles of the Stackable Table to make the headings match. Right now every other `th` is getting `background-color: inherit` which gives them a different background color.

dcmouyard (approved):

PR #520, opened 2021-11-24
Update load_balancer.tf

msglover:
Hi Alex, we got the following error during the terraform validate step in the pipeline:

```
Error: Required attribute is not set
  on load_balancer.tf line 142, in resource "aws_lb_listener" "alb_https":
 142: resource "aws_lb_listener" "alb_https" {

```
It needed the default action block.  I believe I've got it correct in here, but please take a look and let me know if you see any issues.

### DIFF
--- a/terraform/infrastructure/load_balancer.tf
+++ b/terraform/infrastructure/load_balancer.tf
@@ -146,6 +146,11 @@ resource "aws_lb_listener" "alb_https" {
 
   port     = 443
   protocol = "HTTPS"
+  
+  default_action {
+    type = "forward"
+    target_group_arn = aws_lb_target_group.alb_http.arn
+  }
 }
 
 resource "aws_lb_listener_certificate" "alb_https" {


terraform/infrastructure/load_balancer.tf
```diff
@@ -146,6 +146,11 @@ resource "aws_lb_listener" "alb_https" {
 
   port     = 443
   protocol = "HTTPS"
+  
+  default_action {
+    type = "forward"
+    target_group_arn = aws_lb_target_group.alb_http.arn
```
gustavderdrache, line 152: In order to match what we have for Traefik (where requests to unrecognized domains get dropped), I'd prefer if we used the ALB's fixed response default action. This suggestion is for a 204 No Content response, in case you need a positive response for health checks, but this could easily be a 404 or even a 5XX status code.

```suggestion
    type             = "fixed-response"
    
    fixed_response {
      content_type = "text/plain"
      status_code  = 204
    }
```

PR #526, opened 2021-12-02
Two changes to correct errors in the terraform apply step:

msglover:
We saw the following two errors when trying to apply this terraform:

Error: Error modifying LB Listener: ValidationError: An SSL policy cannot be specifed for TCP listeners
	status code: 400, request id: eafb1313-d07b-4b1a-b761-27209e181d69
  on load_balancer.tf line 48, in resource "aws_lb_listener" "https":
  48: resource "aws_lb_listener" "https" {
Error: vpc_id should be set when target type is ip
  on load_balancer.tf line 110, in resource "aws_lb_target_group" "alb_http":
 110: resource "aws_lb_target_group" "alb_http" {


gustavderdrache (approved):
Good catch!

PR #541, opened 2021-12-14
Epad8 1555 akamai to cloudfront no purge

azinck:
Removes the akamai module and disables the purger. Does not put a cloudfront purger in place; will rely on low TTL.


azinck:
Holding off on merging this until we get more clarity from EPA about the timeline for deploying it.

PR #549, opened 2021-12-27
Epad8 1559 dynamic lists duplicate perspectives

nh0713:


services/drupal/web/modules/custom/epa_workflow/epa_workflow.deploy.php
```diff
@@ -215,3 +215,77 @@ function epa_workflow_deploy_0004_fix_nodes_missing_transition_date(&$sandbox) {
...
+      foreach ($tids as $tid) {
+        $node->field_author_names[] = [
+          'target_id' => $tid
+        ];
+      }
+    }
+    else {
+      // Clear new field value if field_authors is empty.
+      $node->field_author_names = [];
+    }
```
azinck, line 287: Do we really need to do all this here @nh0713 ? Can't we just do the $node->save() and let our new presave logic handle it?

PR #551, opened 2021-12-27
[EPAD8-1265] Setting default values for channel and type on regulation nodes.

carsoncho:


azinck:
Please see feedback on the issue.

PR #560, opened 2022-01-07
EPAD8-1578 add wrappers and attributes necessary to support drag and …

azinck:
…drop in our layout


azinck:
@dcmouyard Would you look at this PR? In order to support drag and drop properly in LB we have to have a wrapper around the layout that has the appropriate attributes. If you are staunchly opposed to this div, then we may have to see if there's a way to determine if we're in the editing context and only add it there.

dcmouyard:
@azinck I fixed two typos in the resource directory layout which caused the broken sidebar issue QA found.

PR #579, opened 2022-01-14
Epad8 1579 improve the experience of automatic publishing

nh0713:


services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -647,6 +647,39 @@ function epa_workflow_module_implements_alter(&$implementations, $hook) {
...
+  $node = $form_state->getBuildInfo()['args'][0];
+  if ($node instanceof NodeInterface && !$node->isNew()) {
+    // Loading will always return the "Current revision".
+    /** @var \Drupal\node\NodeInterface $current */
+    $current = \Drupal::entityTypeManager()
+      ->getStorage('node')
+      ->load($node->id());
+
+    // If editing latest revision and has a Publish Date set.
+    if ($node->getRevisionId() == $current->getRevisionId() && isset($node->field_publish_date->value)) {
```
azinck, line 660: I assume (though I'm not totally sure) that the $node object here is the latest revision.  If so, why are we checking to see if it matches the current revision?  We want all of the logic here to apply regardless of whether the current and latest revision are the same, unless I'm missing something.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -647,6 +647,39 @@ function epa_workflow_module_implements_alter(&$implementations, $hook) {
...
+      }
+      // Is in any other draft state.
+      else if ($moderation_state == 'draft' || $moderation_state == 'draft_needs_review') {
+        $message .= '<br>The revision is scheduled to publish on ' . $date . ' at ' . $time . ', but will only do so if this draft is approved. Currently, the draft is NOT approved.';
+        $form['node_revisions_table'][0]['revision']['#context']['message']['#markup'] = $message;
+      }
+      // If the current revision is published and has a Sunset Date set, update Sunset Date help text.
+      else if ($node->status->value == 1 && isset($node->field_expiration_date->value)) {
+        $message .= '<br>This content will automatically sunset, and be unpublished, on ' . $date . ' at ' . $time . '.';
+        $form['node_revisions_table'][0]['revision']['#context']['message']['#markup'] = $message;
```
azinck, line 679: The "current revision" (which is what the published revision is) will not always be the "latest revision".  Consider what will happen if there are forward draft revisions. We need to figure out which row it's in and modify that row; we can't just hard-code index 0. 

A couple of ways you can figure out which row is correct: you can get the vid of the current revision from the radio button form elements.  An alternative way is that the 'revision-current' class is applied only to the row of the current revision, so you can inspect the #attributes array for that class.

azinck: Note that hard-coding the 0-index on the other bits of logic is fine because they always only apply to the latest revision. It's only the sunset date logic that needs to always be applied to the current revision.

azinck: In addition to the obvious test cases, be sure you do some testing with forward revisions.  That is, have a published revision (with sunset date) and create a new draft that has a publish date on it and ensure our messages about it needing to be in draft approved in order to be automatically published are displayed appropriately, and ensure the sunset date message displays appropriately on the published revision.

No published revision, just drafts, and the latest revision has both

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -647,6 +647,39 @@ function epa_workflow_module_implements_alter(&$implementations, $hook) {
...
+
+  $node = $form_state->getBuildInfo()['args'][0];
+  if ($node instanceof NodeInterface && !$node->isNew()) {
+    // Loading will always return the "Current revision".
+    /** @var \Drupal\node\NodeInterface $current */
+    $current = \Drupal::entityTypeManager()
+      ->getStorage('node')
+      ->load($node->id());
+
+    // If editing latest revision and has a Publish Date set.
```
azinck, line 659: We're not in the editing context, so let's remove the reference to editing just for clarity.

azinck (changes requested):

azinck (approved):

azinck (approved):

PR #587, opened 2022-01-25
Epad8 1595 expand paragraph editing permissions

nh0713:


azinck (changes requested):
Feedback left in Jira.

azinck (changes requested):

PR #597, opened 2022-02-14
[EPAD8-1472] add Native Lazy Loading module

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1472

Should just enable the native lazy loading module. Should enable proper lazy loading on images.


azinck:
Merged an older version of this branch to avoid bringing in the event date changes that were irrelevant.

PR #599, opened 2022-02-15
EPAD8-1629 mobile menu fix

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1629

Adjusts how the JS is loaded in so we don't double-bind the menu toggle.

### DIFF
--- a/services/drupal/web/themes/epa_theme/js/src/modules/navigation.js
+++ b/services/drupal/web/themes/epa_theme/js/src/modules/navigation.js
@@ -1,5 +1,3 @@
-import navigation from 'uswds/src/js/components/navigation.js';
-
 export default function() {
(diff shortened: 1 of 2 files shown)


services/drupal/web/themes/epa_theme/source/_patterns/04-layouts/header/_header.scss
```diff
@@ -2,8 +2,6 @@
 // Styles for Header Layout.
 
 .l-header {
-  position: relative;
-  z-index: 2;
```
dcmouyard, line 6: @jbwagner Why are these lines removed? We noticed a new bug during the demo today where the active local task menu item was shown on top of dropdown menus.

dcmouyard (approved):

PR #609, opened 2022-02-24
EPAD8-1622: Adding review deadline date to moderation window if a nod…

rjensen26:
Things that were done:
- [x] Added `field_review_deadline` to all `moderation block` to all `published content`

### DIFF
--- a/services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
+++ b/services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
@@ -259,34 +259,26 @@ function epa_workflow_form_alter(array &$form, FormStateInterface $form_state, $
     // on the moderation block only on published content
     // and the `field_review_deadline` has a value
     if (
-      isset(
-        $node,
-        $node->field_review_deadline,
-        $node->field_review_deadline->date)
+      isset($node)
+      && !$node->field_review_deadline->isEmpty()
       && $node->isPublished()
       )
     {
 
-      $timezone = date_default_timezone_get();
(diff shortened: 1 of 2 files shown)


services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
```diff
@@ -254,6 +254,34 @@ function epa_workflow_form_alter(array &$form, FormStateInterface $form_state, $
...
+
+    // Want to display the `field_review_deadline` date
+    // on the moderation block only on published content.
+    if (isset($node) && $node->isPublished()) {
+
+      $timezone = date_default_timezone_get();
+      $date_formatter = \Drupal::service('date.formatter');
+
+      $review_deadline = $date_formatter->format(
+        $node->field_review_deadline->date->getTimestamp(), 'custom', 'F d, Y h:i a T', $timezone
```
azinck, line 266: Let's guard against the field_review_deadline possibly not being set.  It really should always be set for published nodes, but if some policies fail to occur properly then it's possible for it to be unset and we don't want to throw an error, we should just not display the field.

azinck (changes requested):
This looks good I think -- just needs a guard in the event the date value is missing.

services/drupal/web/modules/custom/epa_workflow/epa_workflow.module
rjensen26: Added a check for `field_review_deadline` & `field_review_deadline->date`

PR #628, opened 2022-03-11
EPAD8-1574 js layout shift

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1574


services/drupal/web/themes/epa_theme/source/_patterns/03-uswds/banner/banner.twig
```diff
@@ -18,7 +18,7 @@
     </div>
   </header>
   {% if has_expanded %}
-    <div class="usa-banner__content usa-accordion__content" id="gov-banner">
+    <div class="usa-banner__content usa-accordion__content" id="gov-banner" hidden>
```
dcmouyard, line 21: @jbwagner We shouldn’t use the `hidden` attribute since that will hide the content from screen readers. Can you switch to using another custom attribute or class?

dcmouyard (changes requested):

services/drupal/web/themes/epa_theme/source/_patterns/03-uswds/banner/banner.twig
jbwagner: @dcmouyard that's what the JS code was using to toggle visibility initially. Which wouldn't matter I suppose when it's initially loaded, just complicates it for load (of course).

dcmouyard (approved):

PR #642, opened 2022-03-25
[EPAD8-1647] add aria-current attribute to current menu item

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1647


dcmouyard (changes requested):
@jbwagner Works great! Can you also add `aria-current="page"` to the button group component? (`source/_patterns/05-components/button-group/button-group.twig`)

dcmouyard (approved):

PR #643, opened 2022-03-28
[EPAD8-1625] add multilingual fonts to local

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1625


services/drupal/web/themes/epa_theme/epa_theme.libraries.yml
```diff
@@ -18,42 +18,10 @@ global:
...
-        preprocess: false,
-        type: external,
-      }
-font_traditional_chinese:
-  css:
-    theme:
-      //fonts.googleapis.com/css2?family=Noto+Sans+TC:wght@400;700&display=swap: {
-        preprocess: false,
-        type: external,
-      }
```
dcmouyard, line 56: @jbwagner Can you add these libraries back? Have them point to new CSS files rather than to the google hosted ones. This will allow us to only load specific languages in the future (which is on EPA’s roadmap.)

dcmouyard (changes requested):

services/drupal/web/themes/epa_theme/epa_theme.libraries.yml
jbwagner: @dcmouyard added them back. Per our Slack convo, I kept the preloads I added for now until we solve for the future case of only loading exactly what we need.

PR #644, opened 2022-03-29
EPAD8-1688: Fix missing revisions tab.

slucero:
Apply patch to the Group module for this issue to fix the missing revisions tab after updating core to 9.3.x: https://www.drupal.org/project/group/issues/3256998.


azinck (changes requested):
Unfortunately this patch doesn't completely fix the issue.  See comments in Jira.  Sending back for more investigation.

PR #658, opened 2022-04-07
[EPAD8-1630] update external links with new allowed

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1630


dcmouyard (approved):

services/drupal/web/themes/epa_theme/js/src/external-links.es6.js
```diff
@@ -21,9 +29,15 @@ import Drupal from 'drupal';
...
           external = false;
         }
-        if (allowedDomains.includes(linkElement.host)) {
-          external = false;
-        }
+
+        allowedDomains.forEach(domain => {
+          if (
+            linkElement.host === domain ||
+            linkElement.host.endsWith(domain)
```
azinck, line 36: @jbwagner This isn't a good way to check.  Consider a domain like `pwnedclu-in.org` -- it'll match this check.  We need to check that the linked domain either exactly matches, or that it ends with, for instance `.clu-in.org` (the leading dot ensuring that we're only matching subdomains).

jbwagner: @azinck added a leading `.` character to the check to make sure it finds accurate subdomains

azinck (approved):

PR #659, opened 2022-04-07
[EPAD8-1642] update date position in footers

jbwagner:
JIRA: https://forumone.atlassian.net/browse/EPAD8-1642


dcmouyard (changes requested):
@jbwagner Since we’re moving `footer__last-updated` to inside `l-page__footer` it’s no longer part of the `footer` component. We should rename it to `l-page__footer-last-updated`. We should also wrap the Contact Us info in `l-page__footer-contact`.

The CSS should then be updated to target those new classes instead of generic `<span>` elements.

PR #666, opened 2022-04-12
Update redirect_301.map

caedstrom:
removed dfe global redirect to safechoice


azinck:
@caedstrom  Do you really wish to remove this?  This will also remove the redirection of any requests to sub-paths of /dfe (like `/dfe/about` or anything like that)

azinck:
Going to do this separately.

PR #688, opened 2022-04-29
Possible file that can be deleted

rjensen26:


rjensen26:
@gustavderdrache , Added you on here just to verify if you had any thoughts of removing the specified files.

azinck:
Per Alex, we cannot merge this PR until we've collaborated with GDIT to decommission Traefik.  Quote: "just need to coordinate with GDIT when we can remove it from ECR and stuff - it's possible they're still building it in gitlab so it'll break builds if we don't give them a heads up"

PR #730, opened 2022-06-27
[EPAD8-1734] - adds user status and last access to the user profile page 

pzh7000:


services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
```
azinck, line 9: You don't seem to be using this anywhere.

azinck (approved):

PR #747, opened 2022-08-01
EPAD8-1545: Before/After Swipe component

pzh7000:


services/drupal/web/modules/custom/view_mode_select_by_token/src/Plugin/Field/FieldFormatter/RenderedWithSelect.php
```diff
@@ -0,0 +1,97 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\entity_browser\Plugin\EntityBrowser\FieldWidgetDisplay\RenderedEntity;
+
+
+/**
+ * Plugin implementation of the 'Render entity to Markdown' formatter.
+ *
+ * @FieldFormatter(
+ *   id = "view_mode_select_by_token_rendered_with_select",
+ *   label = @Translation("Rendered With Select"),
```
azinck, line 16: I'm wondering if we can come up with another label for this. "Rendered With Select" doesn't really mean much to me. Maybe something like "Render with variable view mode", or "Render with dynamic view mode"?  Any thoughts on that? Not the biggest deal, just trying to make this be something the general public will understand.

services/drupal/web/modules/custom/view_mode_select_by_token/src/Plugin/Field/FieldFormatter/RenderedWithSelect.php
```diff
@@ -0,0 +1,97 @@
...
+
+      // Add a resource attribute to set the mapping property's value to the
+      // entity's url. Since we don't know what the markup of the entity will
+      // be, we shouldn't rely on it for structured data such as RDFa.
+      if (!empty($items[$delta]->_attributes) && !$entity->isNew() && $entity->hasLinkTemplate('canonical')) {
+        $items[$delta]->_attributes += ['resource' => $entity->toUrl()->toString()];
+      }
+    }
+
+    return $elements;
```
azinck, line 94: Basically all of this can be discarded; it's just a copy from the parent. You should be able to just set the view mode setting then call the parent method:
```suggestion
$this->setSetting('view_mode', $view_mode);
return parent::viewElements($items, $langcode);
```

services/drupal/web/modules/custom/view_mode_select_by_token/view_mode_select_by_token.install
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @file
+ * Install, update and uninstall functions for the View Mode Select by Token module.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function view_mode_select_by_token_install() {
+  \Drupal::messenger()->addStatus(__FUNCTION__);
```
azinck, line 12: I've not really seen this...is this a standard you found somewhere?

services/drupal/web/modules/custom/view_mode_select_by_token/config/schema/view_mode_select_by_token.schema.yml
```diff
@@ -0,0 +1,7 @@
+field.formatter.settings.epa_core_rendered_with_select:
+  type: mapping
+  label: 'Rendered With Select'
+  mapping:
+    text_length:
+      type: string
+      label: 'View Mode'
```
azinck, line 7: I applaud you for including the schema file but I'd have expected the key for this to be `field.formatter.settings.view_mode_select_by_token_rendered_with_select` and the mapping to reference `view_mode` (rather than `text_length`) along with an entry for `link` since we're still preserving that setting.

services/drupal/web/modules/custom/view_mode_select_by_token/view_mode_select_by_token.install
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function view_mode_select_by_token_uninstall() {
```
azinck, line 18: Same comment as for the install hook.

services/drupal/web/modules/custom/view_mode_select_by_token/view_mode_select_by_token.install
```diff
@@ -0,0 +1,96 @@
...
+  \Drupal::messenger()->addStatus(__FUNCTION__);
+}
+
+/**
+ * Implements hook_schema().
+ */
+function view_mode_select_by_token_schema() {
```
azinck, line 25: This module doesn't store any data in a table so should not have this hook implemented.

services/drupal/web/modules/custom/view_mode_select_by_token/view_mode_select_by_token.install
```diff
@@ -0,0 +1,96 @@
...
+    ],
+  ];
+
+  return $schema;
+}
+
+/**
+ * Implements hook_requirements().
+ */
+function view_mode_select_by_token_requirements($phase) {
```
azinck, line 83: I'm guessing this was copied from somewhere. We don't have any special requirements. The current implementation here seems to be designed to randomly show an INFO or WARNING, each half the time. Let's delete this.

azinck (changes requested):

dcmouyard (approved):

PR #752, opened 2022-08-16
EPAD8-1807 remove all akamai-related variables and parameters from te…

azinck:
…rraform


azinck:
@gustavderdrache Can you take a look at this and validate that I did this correctly? I'm also curious if you have guidance on the best way to deploy this.  If we just deploy it with these variables/secrets/etc. removed, will it complain and throw any errors? Do we or GDIT need to do anything in our respective AWS accounts to either adjust or remove things in order for this to deploy cleanly?

gustavderdrache (approved):
This will impact our AWS infrastructure (specifically the non-`main` branches) in the following way: new task launches will fail because they will be unable to find the Akamai parameters that the task definition says they should bind to. The easiest fix, once `main` is confirmed working, is to apply this patch to the other branches so they can continue to deploy cleanly.

The impacts to GDIT's deployment are lower because I believe they only deploy one copy of the site per infrastructure. We will want to notify them ahead of time, however, as the production Terraform will be changing and that requires a manual review before they synchronize it.

PR #755, opened 2022-08-17
Update OneEPA_template.R

elinork:
Removing relative link to js/pattern-lab-head-script.js as recommended over email


cherrypj:
Thanks @elinork !

PR #757, opened 2022-08-19
[EPAD8-1698] - adding comoser-merge-plugin and including webform comp…

pzh7000:
…oser.libraries.json to be handled by main composer file


azinck:
This was based on the wrong branch so I'm closing it.

PR #767, opened 2022-09-09
[EPAD8-1808] - adds add_to_cal_description field to events, creates n…

pzh7000:
…ew token and processes token for addtocal module


services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -255,6 +256,10 @@ function epa_core_token_info() {
     'name' => t('Term description'),
     'description' => t('Default term description.'),
   ];
+  $add_to_cal_description = [
```
azinck, line 259: This needs to be removed @pzh7000 since we're not using tokens anymore.

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +378,43 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+  $node = \Drupal::routeMatch()->getParameter('node');
+  if ($node->bundle() == 'event') {
+    $cal_description = $node->get('field_add_to_cal_description')->value;
+    $default_description = $node->get('field_description')->value;
+    if (!empty($cal_description)) {
+      $description = $cal_description;
+    } elseif (!empty($default_description)) {
+      $description = $default_description;
+    }
+    $event_type = Term::load($node->get('field_event_type')->target_id)->getName();
```
azinck, line 398: You don't need to go back and change this, but I believe you could just do: `$node->field_event_type->entity->name` @pzh7000

services/drupal/config/sync/core.entity_view_display.node.event.default.yml
```diff
@@ -58,8 +59,8 @@ content:
         value: field_location
         tokenized: ''
       description:
-        value: field_description
-        tokenized: ''
+        value: '0'
+        tokenized: '[node:add-to-cal-description]'
```
azinck, line 63: Let's ditch this particular bit of configuration since the token no longer exists.  Let's replace it with text that says something like "This value is dynamically generated in epa_core_addtocal_alter()" to make it clear to future devs what's going on here.

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
 }
 
+function epa_core_addtocal_alter(array &$overridable_items) {
+  $overridable_items['override_status'] = TRUE;
+  $node = \Drupal::routeMatch()->getParameter('node');
+  if ($node->bundle() == 'event') {
+    $cal_description = $node->get('field_add_to_cal_description')->value;
+    $default_description = $node->get('field_description')->value;
+    if (!empty($cal_description)) {
+      $description = "$cal_description" . "\n";
```
azinck, line 390: Please clean this up -- should just be `$cal_description . "\n"`...no reason to wrap the var in extra double-quotes.  But, better yet, let's not put a line break at the end of all these.  Just do `$description = $cal_description;` here and we'll add the line breaks between the elements all at once, at the end.

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    } elseif (!empty($default_description)) {
+      $description = "$default_description" . "\n";
```
azinck, line 392: Same issue as above with funky concatenation and double-quotes, but let's just drop the concatenation and \n entirely as we'll add the line breaks at the end.

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    }
+    $event_type = Term::load($node->get('field_event_type')->target_id)->getName();
+    if (!empty($event_type)) {
+      $event_type_full = "Event Type: $event_type" . "\n";
```
azinck, line 396: Drop the line break here, we'll add it at the end. (also, FYI there's no reason to separate the 2 strings here and concatenate them, it could have just been a single string)

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    }
+    $event_host = $node->get('field_event_host')->value;
+    if (!empty($event_host)) {
+      $event_host_full = "Event host: $event_host" . "\n";
```
azinck, line 400: Again, please drop the line break here, we'll add it at the end.

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    }
+    $links = $node->get('field_links')->getValue();
+    $links_string = "Links:" . "\n";
```
azinck, line 403: There's no need to have these be separate strings you're concatenating. Just make it a single string: `"Links:\n"`

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    foreach ($links as $link) {
+      $text = $link['title'];
+      $url = $link['uri'];
```
azinck, line 406: Unfortunately this isn't going to work.  The uri can hold an actual link, but can also hold a reference to another node.  When that's the case, you get something output that looks like this: `entity:node/276985`.

We need to do proper link generation...reference the field formatter for this field to get some idea of how to do that: `\Drupal\link\Plugin\Field\FieldFormatter\LinkFormatter::viewElements`

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+      $links_string .= "$text" . ": " . $url . "\n";
```
azinck, line 407: No need for the extra quotes around `"$text"`.  Also, if you're using double-quotes like this, no need for concatenation at all when referencing variables.  `$links_string .= "$text: $url\n";` will work just fine.  That said, explicit concatenation can assist with readability so that's fine

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +374,45 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    }
+    $full_description = $description . $event_type_full . $event_host_full . $links_string;
```
azinck, line 409: It's going to be a bit cleaner/clearer to just insert all the line breaks here at once.  And I'd like to see some separation between each chunk (not just putting it on a new line, but adding an additional empty line to separate each bit).  So that would yield `$full_description = $description . "\n\n" . $event_type_full . "\n\n" . $event_host_full . "\n\n" . $links_string;` (you'll see I went with explicit concatenation there for readability, but that's just preference)

azinck (changes requested):

services/drupal/web/modules/custom/epa_core/epa_core.module
arydarowski: I made some updates, there is probably a better way to go about putting the internal link together, suggestions welcome!

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +376,51 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    $links = $node->get('field_links')->getValue();
+    $links_string = "Links:\n";
+    foreach ($links as $link) {
+      $text = $link['title'];
+      $url = $link['uri'];
+      if (!UrlHelper::isExternal($url)){
+        $nid = Url::fromUri($url)->getRouteParameters()['node'];
+        $alias = \Drupal::service('path_alias.manager')->getAliasByPath('/node/'.$nid);
+        $host = \Drupal::request()->getHost();
+        $url = "https://".$host.$alias;
```
arydarowski, line 413: not sure if it is the host value we want here, on my local it resolves to https://epa.ddev.site/epalibraries/headquarters-repository-library

services/drupal/web/modules/custom/epa_core/epa_core.module
```diff
@@ -373,13 +376,51 @@ function epa_core_tokens($type, $tokens, array $data, array $options, Bubbleable
...
+    $event_host = $node->get('field_event_host')->value;
+    if (!empty($event_host)) {
+      $event_host_full = "Event host: $event_host";
+    }
+    $links = $node->get('field_links')->getValue();
+    $links_string = "Links:\n";
+    foreach ($links as $link) {
+      $text = $link['title'];
+      $url = $link['uri'];
+      if (!UrlHelper::isExternal($url)){
```
arydarowski, line 409: I'm still thinking over this check here, perhaps a different check will be better but this is what it is for now

PR #769, opened 2022-09-12
EPAD8-1771: Fix font links in standalone template

dcmouyard:


sarahprops:
@dcmouyard 1771 fix is in this PR

PR #770, opened 2022-09-12
[EPAD8-1804] - adds node id to entity moderation form

pzh7000:


services/drupal/web/themes/epa_theme/includes/form.inc
```diff
@@ -30,6 +30,51 @@ function epa_theme_preprocess_form(&$vars) {
   }
 }
 
+/**
+ * Implements template_preprocess_form().
+ */
+function epa_theme_preprocess_entity_moderation_form(&$vars) {
+  $vars['testing'] = 'preprocess';
```
azinck, line 37: @dcmouyard Please remove this.

services/drupal/web/themes/epa_theme/templates/form/entity-moderation-form.html.twig
```diff
@@ -2,13 +2,57 @@
...
-  <li class="entity-moderation-form__item">{{ form.review_deadline }}</li>
-  <li class="entity-moderation-form__item">{{ form.new_state }}</li>
-  <li class="entity-moderation-form__item">{{ form.revision_log }}</li>
-  <li class="entity-moderation-form__item">{{ form.workflow_508_compliant }}</li>
-  <li class="entity-moderation-form__item">{{ form.submit }}</li>
-</ul>
-{{ form|without('current', 'revision_author', 'review_deadline', 'new_state', 'revision_log', 'submit', 'workflow_508_compliant', 'node_edit_link') }}
+<div class="box box--special">
+  <div class="box__content">
+    {% if form.nid %}
```
azinck, line 7: I'm surprised we couldn't override the form field templates directly rather than doing this workaround.  Did you look into that @dcmouyard ?

azinck (changes requested):

services/drupal/web/themes/epa_theme/templates/form/entity-moderation-form.html.twig
dcmouyard: @azinck There aren’t any template suggestions for form items, so I went with this approach instead of creating new template suggestions.

Another approach would be to pipe these through the correct theme functions in `epa_workflow.module`, but I had already done the current approach before I learned where these items were set.

azinck: I can live with this, thanks.

azinck (approved):

PR #773, opened 2022-09-14
[EPAD8-1656] - installs and enables metatag module, sets up some test…

azinck:
… configuration and creates new metatag field for web areas


services/drupal/config/sync/metatag.settings.yml
```diff
@@ -0,0 +1,10 @@
+entity_type_groups:
```
azinck, line 1: @pzh7000 We don't really want this file.  This exists because you enabled all of the checkboxes for the tags for this content type, but that's unnecessary.  Let's remove it.

services/drupal/config/sync/metatag.metatag_defaults.node__public_notice.yml
```diff
@@ -6,3 +6,10 @@ id: node__public_notice
 label: 'Content: Public Notice'
 tags:
   title: 'Public Notice: [node:title] | [site:name]'
+  schema_special_announcement_category: 'https://www.wikidata.org/wiki/Q81068910'
+  schema_special_announcement_type: SpecialAnnouncement
+  schema_special_announcement_spatial_coverage: 'a:2:{s:5:"@type";s:7:"Country";s:4:"name";s:2:"US";}'
+  schema_special_announcement_date_posted: '[node:created:html_datetime].'
```
azinck, line 12: @pzh7000 Per the AC, they want the last published date here, which is maintained in its own field.

PR #848, opened 2022-11-01
rerolling patches for content moderation notifications group, origina…

arydarowski:
…l revision author and flags support


services/drupal/composer.json
```diff
@@ -231,7 +231,7 @@
         "drupal/config_filter": "^1.8",
         "drupal/config_ignore": "^2.3",
         "drupal/config_split": "~1.5",
-        "drupal/content_moderation_notifications": "dev-3.x#7ffa936",
+        "drupal/content_moderation_notifications": "dev-3.x",
```
azinck, line 234: I typically prefer pinning the commit in the dependency specification when tracking dev branches to avoid accidental updates to newer untested versions of the software.  So if we need to update to a newer hash let's bake that in here.

services/drupal/composer.lock
```diff
@@ -1675,16 +1675,16 @@
         },
         {
             "name": "doctrine/collections",
-            "version": "1.7.3",
+            "version": "1.8.0",
```
azinck, line 1678: I'm very suspicious of these unrelated dependency updates.  I wonder how they crept in here.  Just running `composer require drupal/content_moderation_notifications:dev-3.x` doesn't lead to these dependencies changing. Any thoughts on what might be happening here?  This comment is relevant for this package as well as several others like phpdocumentor/type-resolver, doctrine/event-manager, and others.

services/drupal/composer.patches.json
```diff
@@ -77,9 +77,9 @@
       "Do not reimport existing entities": "https://www.drupal.org/files/issues/2020-11-06/default_content-existing-entities-2698425-1-x.patch"
     },
     "drupal/content_moderation_notifications": {
-        "Group Entity Support": "https://www.drupal.org/files/issues/2020-02-17/content_moderation_notifications-3015275-35.patch",
-        "Add option to email Revision Author": "patches/content-moderation-notifications-3133940-2-rolled-to-apply-with-3015275.patch",
-        "Flag module support": "patches/content-moderation-notifications-3316633-flag-support-rolled-to-apply-with-3015275-and-3133940.patch"
+      "Group Entity support": "patches/content_moderation_notifications-with-issues-3015276-comment-14511793-patch-applied.patch",
+      "Add option to email Revision Author": "patches/content_moderation_notifications_issues-3133940-comment-14693459-patch-over-issues-3015276-comment-14511793-patch.patch",
```
azinck, line 81: Let's clean up the naming of these patches.  I'd like them to be named content_moderation_notifications-[issue ID of the issue the patch is meant to fix]-[comment # (not ID, just the number of the comment on the issue) that contains the patch that this patch was based on]-rolled-to-apply-with-[issue IDs of each previously-applied-patch].patch.  

For example: `content-moderation-notifications-3133940-2-rolled-to-apply-with-3015275.patch`

services/drupal/patches/content_moderation_notifications-with-issues-3015276-comment-14511793-patch-applied.patch
```diff
@@ -0,0 +1,420 @@
+diff --git a/config/schema/content_moderation_notifications.schema.yml b/config/schema/content_moderation_notifications.schema.yml
```
azinck, line 1: This patch file is not in use by us so should be removed.

PR #444, opened 2021-09-20
Update metatag.metatag_defaults.front.yml

cherrypj:
https://forumone.atlassian.net/browse/EPAD8-1455

For some reason, the global defaults do not seem to apply to the front page, so we're adding these for the front page (in order to clear errors that FB and Twitter debuggers throw). I did not know how to generate a UUID for the Spanish version of metatag.metatag_defaults.front.yml

PR #887, opened 2022-12-06
EPAD8-1903 EPA Claro color contrast improvement for layout builder

Purple-SarahProper:
this issue is caused by the jquery ui theme.css setting any links under .ui-widget-content to be #333 and the layout builder .css setting the background color on list items links to have a background color of #444. 

this PR sets the link color to white and lightly updates the styling of the list items. 

#fff / #444 passes color contrast requirements. 
Screen Shot 2022-12-05 at 11 19 36 PM


Purple-SarahProper:
sarah signed into the wrong account will reopen with the right permissions!

PR #926, opened 2023-01-06
Generate PL URLs for Diffy Visual Regression testing

azinck:


azinck:
@dcmouyard Please take a look at this locally since the Github diff viewer can't handle the size of the package-lock.json diff.  I had to add a couple of npm packages and it made a large number of changes to the lock file for reasons I'm not entirely clear about.  It even changed the lockfileVersion property from 2 to 1 which to me implies I generated this using a different version of npm than you were using.  To be clear, I'm running npm commands through ddev like: `ddev gesso npm install x`.  Are you running it locally? Does that account for the difference?

PR #928, opened 2023-01-09
Epad8 1942 adjust bulk publish

arydarowski:


services/drupal/web/modules/custom/epa_workflow/src/Plugin/Action/EPABulkRepublishAction.php
```diff
@@ -0,0 +1,211 @@
...
+   */
+  public function access($object,  AccountInterface $account = NULL, $return_as_object = FALSE) {
+    if (!$object->isDefaultRevision()) {
+      $object = \Drupal::entityTypeManager()->getStorage($object->getEntityTypeId())->load($object->id());
+    }
+    if (!$object || !$object instanceof ContentEntityInterface) {
+      $result = AccessResult::forbidden('Not a valid entity.');
+      return $return_as_object ? $result : $result->isAllowed();
+    }
+    $object = $this->loadLatestRevision($object);
```
azinck, line 149: I'm pretty sure this is a mistake, and is causing the bug found here: https://forumone.atlassian.net/browse/EPAD8-1942?focusedCommentId=511987 .  The whole point of this new actionis that we want to be acting on the default/current revision, not the latest revision.

azinck (changes requested):

azinck (approved):

PR #932, opened 2023-01-10
[EPAD8-1962] field_press_office cardinality update

carsoncho:


services/drupal/web/modules/custom/epa_core/epa_core.install
```diff
@@ -93,3 +93,18 @@ function _epa_core_create_news_release_type_term() {
   $term->field_term_days_til_review = 730;
   $term->save();
 }
+
+/**
+ * As part of https://forumone.atlassian.net/browse/EPAD8-1962 to set cardinality
+ * of the Press Office field to 1 we will remove records where delta > 0.
+ */
```
azinck, line 100: Let's be a bit more defensive here and guard against cases where there's a value here that's stored in a delta higher than 0, while not having a 0-delta value.

services/drupal/web/modules/custom/epa_core/epa_core.install
```diff
@@ -93,3 +93,42 @@ function _epa_core_create_news_release_type_term() {
...
+ * than the minimum delta value.
+ */
+function epa_core_update_9004() {
+  $database = \Drupal::database();
+  foreach (['node__field_press_office', 'node_revision__field_press_office'] as $table) {
+    if ($database->schema()->tableExists($table)) {
+      // Query to find all rows with the lowest delta value.
+      // SELECT MIN(delta) FROM {$table}
+      $sub_query = $database->select($table)
+        ->addExpression('MIN(delta)');
```
azinck, line 109: As discussed on Zoom, this subquery will find the lowest delta in the whole table, not on a per-revision-id basis.

PR #961, opened 2023-01-31
EPAD8-1966: CKEditor 5 stylesheet

liquilife:
I have addressed the build errors which was introduced when https://forumone.atlassian.net/browse/EPAD8-1809 was merged with this feature branch. In theory this fix should be proactive, no other branches should recreate the same errors. 

The issue:
Editor-styles.scss uses meta.load-css, which was also using the $with parameter to provide an option $wysiwyg variable. In most cases, this requires setting $wysiwyg to false by default in the referenced stylesheets. 

However, when those stylesheets were referenced elsewhere using @forward we get the build error of using $wysiwyg multiple times from multiple SCSS files, which is not allowed. This happened specifically when core.scss was introduced into this branch. 

The fix:
Since $wysiwyg is used minimally in 3 SCSS files, I've removed the $wysiwyg variables and modified the CSS to target the editor directly. This doesn't seem to be creating any other issues and feels like the fix with the path of least resistance.

JIRA:
https://forumone.atlassian.net/browse/EPAD8-1966


liquilife:
All build errors have been fixed and the use of the $wysiwyg variable is now supported in the CKEditor 5 stylesheet `editor-styles.scss`. Updates have been made to editor-styles.scss to ensure that the editing experience in CKEditor 5 is the same as CKEditor 4. 

View:
https://integration.epa.byf1.dev/node/280285/edit

JIRA:
https://forumone.atlassian.net/browse/EPAD8-1966

PR #962, opened 2023-02-01
EPAD8-1747 switching review deadline value to pull from hardcoded arr…

arydarowski:
…ay values for content type and setting to null if not in the array


services/drupal/web/modules/custom/epa_workflow/src/EPAModeration.php
```diff
@@ -223,6 +223,13 @@ protected function setReviewDeadline($reset = FALSE) {
       $date->add(new \DateInterval("P{$review_period}D"));
       $review_deadline = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
       $this->contentEntityRevision->set('field_review_deadline', $review_deadline);
+      if ($bundle == 'public_notice' && !$this->contentEntityRevision->get('field_notice_sort_date')->isEmpty()) {
+        $sort_date = new DrupalDateTime($this->contentEntityRevision->field_notice_sort_date->value);
```
azinck, line 227: I'm thinking there should be a way to get a DrupalDateTime object out of the field directly rather than creating a new one here and having to set the timezone...but I could be wrong.  Could you maybe look into this?

PR #999, opened 2023-02-22
EPAD8-1972: Remove margin-0 class from paragraph blocks

liquilife:
This update removes the class `margin-0` from paragraph blocks.  This prevents the possibility of block titles and paragraph content sitting flush with the previous paragraph block above it. 

This issue was reported happening in the Layout Builder tool. 

Note: 
This ticket is labeled for `Fast Track`, which means this PR is into `main` instead of `release`. 

View: 
https://integration.epa.byf1.dev/node/280302/latest

JIRA:
https://forumone.atlassian.net/browse/EPAD8-1972

### DIFF
--- a/services/drupal/web/themes/epa_theme/templates/block/block--paragraph.html.twig
+++ b/services/drupal/web/themes/epa_theme/templates/block/block--paragraph.html.twig
@@ -4,4 +4,23 @@
  * Theme override for a paragraph block.
  */
 #}
-{% extends "block--admin.html.twig" %}
+{% set block_content %}
+  {% block block_content %}
+    {{ content }}
+  {% endblock %}
+{% endset %}
+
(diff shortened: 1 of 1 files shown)


liquilife:
This update removes the block wrap around the paragraphs created in the Layout Builder when a user is logged in. This prevents the possibility of `margin-0` causing spacing errors on any environment AND targets blocks only created in the Layout Builder. 

To test, view each block rendered via the Layout Builder and verify that they are not wrapped with `<div class="block margin-0"></div>` while logged in and not logged in. 

View: 
https://integration.epa.byf1.dev/node/280302/latest

JIRA:
https://forumone.atlassian.net/browse/EPAD8-1972

tomealter (approved):

PR #1009, opened 2023-02-23
EPAD8-1978 - before / after swipe with touch

sarahprops:
This PR styles the input range slider so it works better with touch devices. 

`appearance: none` for some reason was making it so that the slider could not be slid on mobile. I've done a fair bit of research, but I am unable to figure out the exact reason. Most likely something to do with the clip path on the parent div? 

I also added a min height to the range slider track in -webkit & -moz browsers to make the touch area in the range track larger (outside of dragging the slider button).


sarahprops:
Made a slight change to the size of the slider thumb to give it a defined height / width that's the same size as our visible slider thumb, so that it works with touch on ios devices.

dcmouyard (approved):

PR #1033, opened 2023-03-10
Epad8 1895 selectively include jquery

azinck:


carsoncho:
Closing this PR as jQuery is still very much part of Drupal 10. There's an initiative to remove jQuery entirely and replace it with vanilla JS. See [Drupal 10 javascript dependency plan](https://www.drupal.org/project/drupal/issues/3238507#comment-14688822) and the [Drupal 11 plan to replace jQuery with vanilla JS
](https://www.drupal.org/project/drupal/issues/3052002).

When that comes to fruition we'll revisit this solution.

PR #1047, opened 2023-03-15
Epad8 2048 utility classes

sarahprops:
This PR creates a new stylesheet, utilities.css, that contains the uswds utility classes and our pattern utility classes. it is loaded theme wide with a weight of 50, which means it's always loaded as the last stylesheet.


dcmouyard (approved):

dcmouyard:
Closing this PR since #1050 negates the need for this fix..

PR #1063, opened 2023-03-22
Epad8 2088 combine stylesheets rebased

azinck:
Also integrates EPAD8-2082-mobile-menu-button


azinck:
@dcmouyard I don't have a great sense for how much testing that caniuse change needs.  Do you feel comfortable committing this right now or do you want more testing?

dcmouyard:
@azinck I’m okay with committing this as is. The updated browser list was on integration when QA tested things.

PR #1079, opened 2023-04-13
EPAD8-2121: Remove explicit versions from most theme libraries

dcmouyard:


carsoncho:
Closing this PR as we've come up with a different solution that will still be caching focused. This would change the requests to these files to at least once a day which could be large amounts of requests we don't necessarily need to make.

PR #1099, opened 2023-05-03
EPAD8-822 preserve facets when changing search term

azinck:


azinck:
More work is needed here...it didn't pass QA.

PR #1115, opened 2023-05-24
[EPAD8-2165] Updating version tag for all epa_theme libraries.

carsoncho:


carsoncho:
@dcmouyard I did not update the global library as it was not using the `VERSION` constant. Do you think I should update it?
See [https://github.com/USEPA/webcms/blob/main/services/drupal/web/themes/epa_theme/epa_theme.libraries.yml#L6](https://github.com/USEPA/webcms/blob/main/services/drupal/web/themes/epa_theme/epa_theme.libraries.yml#L6)

dcmouyard:
@carsoncho Yep, the version for `global` should also be `20230524`.

PR #1135, opened 2023-06-20
EPAD8-1958 add minify to minified files in libraries.yml

sarahprops:
This PR adds minified:true to all already minified files (mostly js, 1 css file) set in epa_theme.libraries.yml. According to the [spec](https://www.drupal.org/docs/develop/theming-drupal/adding-assets-css-js-to-a-drupal-theme-via-librariesyml), js files should also have type set when adding the minified parameter. However, since the defualt 'type' is file and all of our files are hosted locally (not external), I believe just having the minified: true suffices.


dcmouyard:
@carsoncho Please merge in #1130 before this one.

dcmouyard (approved):

carsoncho (approved):

PR #1141, opened 2023-07-03
EPAD8-1917: Private files

sarahprops:
this PR:
- adds the "media accessibility" attribute to media items, and sets it to either public or private
- adds a class to media items, files, documents, and the image gallery based on the media accessibility attribute
- styles private media items and files according to the figma file 
- adds the announcement svg to the epa theme, gives it a private title on the private items 
- adds new private alert site alert + styles it
- adds the js to count the number of private items on a page and incorporates it to the site alert style


sarahprops:
@dan I have the latest version of main merged in, which is off from release. also appears to be a composer merge issue (sigh!) that will need to be fixed.

dcmouyard (approved):

PR #1209, opened 2023-10-21
[EPAD8-2270] Web Area disable first link

carsoncho:
- [ ] Incorporates work from EPAD8-2186 as a starting point
- [ ] Utilizing a hook_post_update_NAME() query to find all first group menu link items, compare against the group's field_homepage node's URL and if they are the same (and as long as the menu item has no children), disables the link.


carsoncho:
Closing this as it's incorporated as part of the FEATURE-page-updates branch and must be rolled out at the same time.

PR #1284, opened 2024-01-29
[EPAD8-2340] Adding post-install-cmd to remove the unneeded web.confi…

carsoncho:
…g file.

### DIFF
--- a/services/drupal/composer.json
+++ b/services/drupal/composer.json
@@ -427,7 +427,8 @@
         ],
         "post-install-cmd": [
             "DrupalProject\\composer\\ScriptHandler::createFilesDir",
-            "@install-codestandards"
+            "@install-codestandards",
+            "rm web/web.config"
         ],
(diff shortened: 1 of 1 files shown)


services/drupal/composer.json
```diff
@@ -427,7 +427,8 @@
         ],
         "post-install-cmd": [
             "DrupalProject\\composer\\ScriptHandler::createFilesDir",
-            "@install-codestandards"
+            "@install-codestandards",
+            "rm web/web.config"
```
carsoncho, line 431: @gustavderdrache Do you know if the web-root will always be "web"? Is me hard-coding "web/web.config" here incorrect?

gustavderdrache: We've hardcoded the `web/` directory in a bunch of other configurations, so you're completely fine hardcoding it here.

gustavderdrache (approved):

PR #1301, opened 2024-02-05
Updating to point to the AWS gallery.

carsoncho:


gustavderdrache (approved):
This looks good. It may not completely address build issues, but this will cut down on our pull count significantly.

PR #1333, opened 2024-02-29
EPAD8-2368: Auto subscribe to node when created or updated.

drfuzetto:
Deploy hook for setting notification_opt_in flag for node author, author of the latest revision, or author of the current revision.


services/drupal/web/modules/custom/epa_core/epa_core.deploy.php
```diff
@@ -433,18 +433,157 @@ function epa_core_deploy_0004_set_card_field_default_values(&$sandbox) {
...
+    ON latest.nid = node.nid
+LEFT JOIN node_revision AS current
+    ON node.vid = current.vid
+WHERE latest.vid IN
+    (SELECT MAX(vid)
+    FROM node_revision
+    GROUP BY  nid);")
+      ->fetchAll();
+
+    $sandbox['total'] = count($nodes);
```
azinck, line 475: I'd encourage you to change the query to a count query to avoid transferring this huge amount of data to PHP then having to do the count in PHP.  Just have the DB count it for you.

services/drupal/web/modules/custom/epa_core/epa_core.deploy.php
```diff
@@ -433,18 +433,157 @@ function epa_core_deploy_0004_set_card_field_default_values(&$sandbox) {
...
+LEFT JOIN node_field_data AS node
+    ON latest.nid = node.nid
+LEFT JOIN node_revision AS current
+    ON node.vid = current.vid
+WHERE latest.vid IN
+    (SELECT MAX(vid)
+    FROM node_revision
+    GROUP BY  nid) LIMIT 500
+        OFFSET " . $sandbox['current'] . ";")
+    ->fetchAll();
```
azinck, line 497: Your query doesn't set an order, so you'll get these records back in a non-deterministic fashion leading to the potential paging problems we spoke about.  I'd recommend ordering by node ID ASC.

services/drupal/web/modules/custom/epa_core/epa_core.deploy.php
```diff
@@ -433,18 +433,157 @@ function epa_core_deploy_0004_set_card_field_default_values(&$sandbox) {
...
+
+  foreach ($nodes as $data) {
+    _epa_core_set_notification_opt_in_flag($data);
+    $sandbox['current']++;
```
azinck, line 501: Rather than tracking the number you've processed it would be better to track the high water mark of the last value of the nid you've processed. This is superior because if nodes are deleted while you're in the process of running these batches a simple "count" could end up skipping records. To explain consider a simple example where you're grabbing 3 nodes at a time from this list of IDs:
- 1
- 2
- 3
- 4
- 5
- 6
- 7

Your first batch processes 1, 2, and 3.  If ID 2 is deleted before your code grabs the 2nd batch, then the list looks like:
- 1
- 3
- 4
- 5
- 6
- 7

 second batch will grab nodes 5,6,7 because your offset of 3 skips the first 3 in the list.  In this scenario you end up skipping ID 4 altogether.

If you, instead, track the highest nid you've processed (ID 3 in this example) then your second request will be for the next 3 IDs that are higher than 3 and it'll correctly return records 4, 5, and 6.

carsoncho:
Changing the base branch to merge this into as we don't want this to go live without the rest of the DANSE work.

services/drupal/web/modules/custom/epa_core/epa_core.deploy.php
azinck: @drfuzetto I'd still make the recommendation that you switch to using a high-water-mark approach to manage pagination and for assessing completion rather than just a simple count-based approach.

carsoncho:
@drfuzetto Can you resolve the conflicts for this so that I may merge this in?

services/drupal/web/modules/custom/epa_core/epa_core.deploy.php
```diff
@@ -448,3 +448,128 @@ function epa_core_deploy_0004_set_card_field_default_values(&$sandbox) {
     }
   }
 }
+
+/**
+ * Setting notification_opt_in flag for node author, author of the latest
+ * revision, or author of the current revision
+ */
+function epa_core_deploy_0055_set_watch_flag(&$sandbox) {
```
carsoncho, line 456: @powpow12 Just a heads up this is the deploy hook I was referring to.

PR #1370, opened 2024-04-22
[EPAD8-2425] Removing patch from elasticsearch as the issue it was trying to solve is no longer relevant.

carsoncho:


carsoncho:
Closing the PR as this was deemed not the issue.

PR #1397, opened 2024-06-10
EPAD8-2464: Update EPA Workflow module to use danse_moderation_notifications.

drfuzetto:


carsoncho:
Closing as we don't need this. This is to try and replicate functionality from content moderation notifications which we don't need given we're moving away from content moderation notifications.

PR #1565, opened 2024-11-05
EPAD8-2440: Remove old content info box code

dcmouyard:


carsoncho:
Going to close this for now as for the foreseeable future we will have the old content info box styles and code for non-beta tester users.

PR #1581, opened 2024-11-18
EPAD8-2655: Add phpmyadmin

gustavderdrache:
This PR adds the `ddev/ddev-phpmyadmin` add-on. For some reason, we had some old code related to it, so I just reinstalled to get the add-on updated.


carsoncho (approved):
Works great locally!

PR #1610, opened 2025-02-11
[EPAD8-2683] snapshot link adjustment

carsoncho:


carsoncho:
This update should resolve the issue we're seeing with file links that were not getting updated to be relative links since their URL was not exactly `https://www.epa.gov/...`. This now puts some additional logic in place to do a regular expression check that should match against any subdomain variation as long as it contains `www.epa.gov`.

PR #1618, opened 2025-03-12
[EPAD8-2686] epa-import command

carsoncho:


carsoncho:
@nguyentom715 @ryanletulle Feel free to check out this branch and test out the database import process included here. Ask discussed yesterday, this utilizes a different approach rather than ddev's default `import-db` command which should hopefully speed up the database import for local development.

The instructions should still remain the same for importing the database:
1. Download the .gz or .tar database dump and place it in the `serivces/drupal/.ddev/db` directory
2. Then run `ddev epa-import` and you will be prompted to select which database dump to import. Type the number that corresponds to the database dump and then any other character on the keyboard and press enter. 

Below is an example of what it looks like and then ~10-15 mins later the database will be imported. Much faster than the 30 mins to an hour or more it's probably taking.
```
{25-03-12 11:46:48}~/Projects/webcms/services/drupal:EPAD8-2686-epa-import-command ✓ ➭ ddev epa-import
1) /var/db/innodb-cold-2023-10-05-en-full-mariadb.tar
2) /var/db/integration-es-2024-09-16.sql.gz
3) /var/db/main-en-2024-11-05.sql.gz
#? 1
You have Selected: /var/db/innodb-cold-2023-10-05-en-full-mariadb.tar (1), Please press any key and press enter.
#? d
```

Let me know if you have any questions or run into any issues. Once you've been able to test and validate it's working for you all we can safely merge this into main.

PR #1629, opened 2025-04-25
Update Environmental Indicator module

cherrypj:
Remove irrelevant dev sites, update beta to stage-www, fix color (d.o. #3456599). Not sure patch for path is working (d.o. #3274102)


cherrypj:
WEBCMS-15

PR #1860, opened 2026-05-11
Fix update:stage:es being skipped on every staging pipeline

jmcerda-epa:
Closes https://jira.epa.gov/browse/MOB-2143

## Description

Removes the needs: deploy:stage:apply-es block from update:stage:es in .gitlab-ci.yml. The dependency on a when: manual upstream was leaving the job blocked and ultimately marked as skipped — optional: true doesn't cover the present-but-unstarted case. The job now relies on stage ordering and remains gated by its existing rules: and manual trigger.

To preserve a safety signal, ci/drush.js now compares the deployed ECS Drush image tag against the current build tag and logs a non-fatal ⚠️ warning if they drift, so operators are alerted when running Drush against a stale deployment.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1094,9 +1094,12 @@ update:stage:en:
 update:stage:es:
   extends: .update
   stage: Update
-  needs:
-    - job: "deploy:stage:apply-es"
-      optional: true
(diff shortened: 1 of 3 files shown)


ryanletulle (approved):
coolio. I will be sure to pull the updates.

PR #1862, opened 2026-05-22
CAE-3916: Set Prisma Cloud scan authentication to username/password

jmcerda-epa:
The Prisma Cloud image scans (Drupal, Nginx, Drush) stopped producing results after the authentication method was changed from API-based username/password authentication to a pre-provisioned PRISMA_TOKEN CI variable. This caused twistcli downloads and scans to silently fail (allow_failure: true masked the errors).

This PR makes all three Prisma scan jobs to use the PRISMA_CI_USERNAME / PRISMA_CI_PASSWORD credentials, which are already configured and working. Each job now:
•  Authenticates via the Prisma Cloud /api/v32.01/authenticate endpoint to obtain a bearer token
•  Downloads twistcli using that token
•  Scans images using --user / --password flags

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -523,13 +523,15 @@ Prisma:Drupal:
         exit 0
       fi
   script:
-    # Download twistcli scanner from Prisma Cloud using pre-provisioned token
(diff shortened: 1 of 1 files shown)


ryanletulle (commented):
You convinced me

cherrypj:
Please proceed

cherrypj:
This work isn't on development; does it need to be?

jmcerda-epa:
It does. I merged it.

cherrypj:
Thank you

PR #1866, opened 2026-05-28
refactor(nginx): extract shared FastCGI params into fastcgi-drupal.conf

jmcerda-epa:
Five RSS endpoint locations in `default.conf` each repeated the same 10 FastCGI parameter lines. Any change to buffer sizes or the FPM upstream address required editing all five locations plus the main PHP block.\n\nConsolidate the shared parameters into a new `fastcgi-drupal.conf` snippet and replace each block with a two-line include. Only `PATH_INFO` and `SCRIPT_FILENAME` differ per location and remain inline.

### DIFF
--- a/services/drupal/Dockerfile
+++ b/services/drupal/Dockerfile
@@ -201,7 +201,10 @@ COPY --from=drupal /var/www/html /var/www/html
 # Copy nginx configuration
 COPY status.conf *.map /etc/nginx/conf.d/
 
-# Copy configuration templates
+# Copy configuration templates and the shared FastCGI parameter snippet.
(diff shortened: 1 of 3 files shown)


cherrypj:
I think this work broke development deployment: unknown "webcms_domain" variable.

cherrypj:
Now that #1875 was merged to dev, what do I do with this PR? It's already on dev.

cherrypj:
@jmcerda-epa Did the deployment fail ("fastcgi_pass" directive is not allowed here in /etc/nginx/conf.d/fastcgi-drupal.conf:49) because this PR wasn't merged?

PR #1867, opened 2026-05-28
fix(nginx): standardize S3 proxy DNS resolver to 169.254.169.253

jmcerda-epa:
The `sites/default/files/widgets`, `sites/default/files`, and `archive` proxy blocks were using `127.0.0.11` (Docker embedded DNS) while the `s3fs-css`/`s3fs-js` block already used `169.254.169.253` (AWS VPC DNS). The Docker resolver is not present in ECS Fargate tasks, so those three blocks fail to resolve S3 hostnames in production. Standardize all `proxy_pass` blocks to the VPC resolver.

### DIFF
--- a/services/drupal/default.conf
+++ b/services/drupal/default.conf
@@ -236,11 +236,11 @@ server {
   }
 
   location ^~ /sites/default/files/widgets/ {
-    resolver 127.0.0.11 valid=30s;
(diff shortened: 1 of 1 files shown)


cherrypj:
I don't know if I can merge this if #1866 is wrong.

PR #1870, opened 2026-05-28
fix(ci): add environment guard and dry-run flag to cleanup-webforms

jmcerda-epa:
The script deletes all webform submissions and webform config from the database with no guard against accidentally running on a non-dev environment — nothing prevented it from targeting stage or production.\n\nAdd an explicit check that `WEBCMS_SITE` is `dev` before proceeding. Add a `--dry-run` flag that prints row counts and the SQL that would run without executing anything. Print record counts before deletion so operators know the scope of what is about to be removed.

### DIFF
--- a/ci/cleanup-webforms.js
+++ b/ci/cleanup-webforms.js
@@ -1,12 +1,54 @@
 // @ts-check
-// One-time script to delete all webform submissions and webforms
-// This resolves config import conflicts when webforms have submissions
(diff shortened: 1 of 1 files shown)


cherrypj:
@jmcerda-epa This is your script. Appreciate the guardrails, but ... do we still need the script?

jmcerda-epa:
I dont think it would be a bad thing to keep it around. its reusable.

PR #1895, opened 2026-06-30
Revert "Feature: MOB-2130, Aligning PHP version for all services"

cherrypj:
Reverts WEBSUP-2130. Updating PHP to all be v8.3.x broke a module dependency (the installed version (8.x-1.9) of drupal/diff has a library that requires PHPv8.2--need to update diff to at least 8.x-1.10).


cherrypj:
@zerihunzcgi I had to revert this change; it caused all kinds of problems. Please be sure to test--I'm unable to at the moment.

zerihunzcgi:
@cherrypj  thanks for that. I just checked drupal/diff module and it has a drupal 10/11 update and that should resolve the php issue you have encountered. I will prioritize and test/upgrade modules that are not php 8.3 compatible.

PR #1899, opened 2026-07-02
WEBCMS-311 Remove patch as it is already merged into 4.0.0 version of…

zerihunzcgi:
… paragraphs_entity_embed module

Closes https://jira.epa.gov/browse/WEBCMS-311
## Description

The patch could not be applied because it was already merged in to paragraphs_entity_embed 4.0.0 . This merge removes the patch.

### DIFF
--- a/services/drupal/composer.patches.json
+++ b/services/drupal/composer.patches.json
@@ -133,10 +133,6 @@
     "drupal/linkit": {
       "Editing links does not remove possible stale data- attributes": "https://www.drupal.org/files/issues/2021-08-09/linkit_rename-stale-data-attributes_3111578-3.patch"
     },
-    "drupal/paragraphs_entity_embed": {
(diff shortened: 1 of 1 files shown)


cherrypj:
Just to clarify, this "add openerParameters to ensure media added within embedded paragraphs is associated with the group": "patches/paragraphs_entity_embed-opener_parameters-ck5.patch" patch was custom and was not actually merged. But we will test if deploying to dev still works without the patch. If it doesn't, then @zerihunzcgi will rewrite the patch.

cherrypj:
As expected, without the patch, media embedded in an embedded paragraph are not associated with a group/web area.

zerihunzcgi:
Thank you for confirming @cherrypj. I will look closely and find out what part of the patch is not incorporated in the most recent update.

cherrypj:
None of the patch is incorporated into the plugin code.

PR #1905, opened 2026-07-13
Push Development into Staging

cherrypj:
Version-20260713. Everything since 12 Mar 2026


cherrypj:
@jmcerda-epa I've pushed a bunch of changes into staging. I'd like to test it with a deployment. I've been keeping some notes about difficulties we might run into:

Missing modules
* Clean out core.extension
* Clean out system.schema

Module renamed
* ckeditor5_embedded_content -> embedded_content

Gesso build is now different

cherrypj:
We also need to run the drush commands in WEBSUP-2141.

PR #1948, opened 2026-07-29
WEBCMS-320 Fix Web Area Association by applying patch, patch modified…

zerihunzcgi:
… js file first, npm compile in module and modify patch for the minified js

Closes https://jira.epa.gov/browse/WEBCMS-320

## Description

To fix the issue where a patch for paragraphs_entity_embed was not working, first patch only the file js/ckeditor5_plugins/embeddedParagraph/src/paragraphsembedui.js excluding the minified code from the patch. The reason for that is, we need to incorporate the changes in the patch in the module and then compile using "npm run build". After that we take the minified javascript file from the modules js/build directory and put it back into the patch file. 

I have also removed changes made in the comment section.


cherrypj:
One note, @zerihunzcgi, it seems like the update to version 4 supports double-click. I've been able to double-click an embedded paragraph to edit it.

PR #1971, opened 2026-08-13
Fix AssertionError: Cannot load the "group" entity with NULL ID when …

cherrypj:
Closes WEBCMS-353

## Description

The error occurs because Group::load($target_id) in epa_web_areas_entity_insert() is receiving NULL instead of an integer/string Group ID. To resolve this issue and keep the module safe against future unexpected entity saves (such as REST API calls, background jobs, or entity cloning), we need to:

- Add defensive checks in epa_web_areas_entity_insert() to make sure $target_id is present and valid before attempting to load the Group entity.
- Safely handle empty groups in epa_web_areas_form_node_form_alter().

### DIFF
--- a/services/drupal/web/modules/custom/epa_web_areas/epa_web_areas.module
+++ b/services/drupal/web/modules/custom/epa_web_areas/epa_web_areas.module
@@ -34,56 +34,89 @@ use Drupal\node\NodeInterface;
  *
  */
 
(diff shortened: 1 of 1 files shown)


cherrypj:
@zerihunzcgi Please review and test. Push to dev if you're satisfied.

zerihunzcgi:
Thanks @cherrypj This looks good. I'll merge this and build with the fix for openerParameters warning issue.

